PR #29, opened 2018-10-19
Removed sample data files and removed loading script from schema.json

komick:


craigmcnally:
removing things that were migrated to enumerations (workflow_status, etc.) makes sense.  I think for some of the other stuff like cost, eresource, etc. it might be better to fix the data to conform to the schema rather than just removing the data altogether

komick:
That data was added to aid Arvind in creating a po_line when ui-orders was still pointing at mod-orders-storage directly.  I don't think it serves a purpose currently.

PR #30, opened 2018-10-22
Modordstor 10

komick:


src/test/java/org/folio/rest/impl/POLineTest.java
```diff
@@ -0,0 +1,222 @@
...
+
+    try {
+      // Run this test in embedded postgres mode
+      // IMPORTANT: Later we will initialize the schema by calling the tenant interface.
+      PostgresClient.setIsEmbedded(true);
+      PostgresClient.getInstance(vertx).startEmbeddedPostgres();
+      PostgresClient.getInstance(vertx).dropCreateDatabase(TENANT_NAME + "_" + PomReader.INSTANCE.getModuleName());
+
+    } catch (Exception e) {
+      e.printStackTrace();
```
craigmcnally: Please use logger

komick: Changes made to use logger in both tests

PR #43, opened 2018-12-11
MODORDSTOR19 - added FundDistributionAPI implementation

grabsefx:
Implemented endpoints for processing FundDistribution entity.
* As you see I have removed multiple code duplication from existing unit tests and placed common logic to a single place to src\test\java\org\folio\rest\impl\OrdersStorageTest.java
* According to https://issues.folio.org/browse/MODORDSTOR-18 endpoint for FundDistributionAPI was changed to /orders-storage/fund_distribution
* Current pull request **should be merged together** with https://github.com/folio-org/mod-orders/pull/40


descriptors/ModuleDescriptor-template.json
```diff
@@ -707,6 +738,43 @@
         "eresource.item.delete"
       ]
     },
+    {
+      "permissionName" : "orders_storage_fund_distribution.collection.get",
```
piotr-kalashuk: Please try to follow [Permissions Naming conventions](https://dev.folio.org/guidelines/naming-conventions/#permissions)
> The naming scheme is a faceted dot-separated string, with the delimited terms as dash-separated words. The first portion is the exact name of the responsible module (back-end modules drop the mod- prefix).

ramls/fund_distribution.raml
```diff
@@ -22,7 +22,7 @@ resourceTypes:
     collection-item: !include raml-util/rtypes/item-collection.raml
 
 
-/fund_distribution:
+/orders-storage/fund_distribution:
```
piotr-kalashuk: Along with adding `orders-storage` prefix the approach suggested in [MODORDSTOR-18](https://issues.folio.org/browse/MODORDSTOR-18) is to use plural form for the collection resources. So it makes sense to use `/orders-storage/fund_distributions`

descriptors/ModuleDescriptor-template.json
grabsefx: Fixed according to naming conventions

ramls/fund_distribution.raml
grabsefx: Changed to /orders-storage/fund_distribution**s**  according to REST naming conventions

PR #51, opened 2018-12-27
MODORDSTOR-31 Receiving history schema and API

javalkar-varun:
## Purpose

This API is required because consumers of this API will need data from 3 tables: pieces, po_line, and details. So, instead of querying each table separately, this GET API will query these 3 tables and consolidate into a View called `receiving_history_view` which can be leveraged by the business logic.
https://issues.folio.org/browse/MODORDSTOR-31

This branch will have build failures until folio-org/acq-models#61 is merged

## Approach

Create a schema - https://github.com/folio-org/acq-models/pull/61
Create a View by joining pieces, po_line and details table
Create API which will read data from the View created above.

## Learning

https://www.postgresql.org/docs/9.2/sql-createview.html
https://www.postgresql.org/docs/9.5/functions-json.html

@folio-org/acquisitions


.gitmodules
```diff
@@ -5,4 +5,4 @@
 [submodule "ramls/acq-models"]
 	path = ramls/acq-models
 	url = https://github.com/folio-org/acq-models.git
-        branch = master
+        branch = MODORDSTOR-31-receiving_history
```
KVupp: Revert these changes to point to the master

src/test/resources/pieces_for_view.sample
```diff
@@ -0,0 +1,10 @@
+{
```
KVupp: we can use the existing sample files already available instead of duplicating them , it will be easier to maintain

src/test/resources/po_line_for_view.sample
```diff
@@ -0,0 +1,58 @@
+{
```
KVupp: use existing sample file

src/test/java/org/folio/rest/impl/ReceivingHistoryTest.java
```diff
@@ -0,0 +1,143 @@
...
+  
+  void verifyViewCollection(String endpoint) {
+    // Verify that there are no existing records in View
+    getViewData(endpoint).then()
+      .log().all()
+      .statusCode(200)
+      .body("total_records", equalTo(0));
+  }
+
+  void verifyReceivingHistoryViewCreated(String endpoint) {
```
KVupp: This method can be combined with the above method if we can take the expected value as parameter, again just for maintainability
ex: verifyViewCollection(String endpoint, int expectedCount)

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -64,7 +64,12 @@
       "run": "after",
       "snippet": "COPY po_line (id, jsonb) FROM 'data/poLines.data' ENCODING 'UTF8' DELIMITER E'\t';",
       "fromModuleVersion": "1.0"
-    }
+    },
+    {
+      "run": "after",
+      "snippet": "CREATE OR REPLACE VIEW receiving_history_view AS SELECT pieces.id AS id, json_build_object('id', pieces.jsonb->>'id', 'po_line_number', po_line.jsonb->>'po_line_number', 'title', po_line.jsonb->>'title', 'receiving_note', details.jsonb->>'receiving_note') AS jsonb FROM pieces LEFT OUTER JOIN po_line ON pieces.jsonb->>'id' = po_line.jsonb->>'id' LEFT OUTER JOIN details ON details.jsonb->>'po_line_id' = po_line.jsonb->>'id';",
```
KVupp: Shouldn't the remaining fields in the receiving history schema like caption, comment, supplement, po_line_id etc.. be handled here?

javalkar-varun: Good catch, I have added all remaining fields.

javalkar-varun:
@craigmcnally - This PR is now open for review.

src/test/java/org/folio/rest/impl/ReceivingHistoryTest.java
```diff
@@ -0,0 +1,135 @@
...
+    .statusCode(200)
+    .body("total_records", equalTo(1));
+  }
+  
+  void verifyViewCollection(String endpoint, int expectedCount) {
+    // Verify that there are no existing records in View
+    getViewData(endpoint).then()
+      .log().all()
+      .statusCode(200)
+      .body("total_records", equalTo(expectedCount));
```
craigmcnally: Ideally we'd have more assertions here...

craigmcnally: Also, I think it would be nice to have more data in the view... maybe spanning multiple PO_lines?

piotr-kalashuk:
Please also update `ModuleDescriptor-template.json` adding new interface with handlers and permissions

ramls/receiving_history.raml
```diff
@@ -0,0 +1,32 @@
...
+
+traits:
+    pageable:  !include raml-util/traits/pageable.raml
+    searchable: !include raml-util/traits/searchable.raml
+    language: !include raml-util/traits/language.raml
+
+resourceTypes:
+    collection-get: !include raml-util/rtypes/collection-get.raml
+
+/orders-storage/receiving_history:
```
piotr-kalashuk: As discussed in the Slack, we are going to start to follow [FOLIO Naming conventions](https://dev.folio.org/guidelines/naming-conventions/#api-endpoints). So probably it makes sense to start applying them at least for the new changes:
> Endpoints use dash-separated strings, with URI parameters as camelCase.

descriptors/ModuleDescriptor-template.json
```diff
@@ -410,6 +410,17 @@
...
       ]
     },
+    {
+      "id": "orders-storage.receiving_history",
+      "version": "1.0",
+      "handlers": [
+        {
+          "methods": ["GET"],
+          "pathPattern": "/orders-storage/receiving_history",
+          "permissionsRequired": ["orders-storage.receiving_history.collection.get"]
```
piotr-kalashuk: Please also define permission definition in the "permissionSets" section.

piotr-kalashuk: As discussed in the Slack, we are going to start to follow [FOLIO Naming conventions](https://dev.folio.org/guidelines/naming-conventions/#permissions) as much as we can. So probably it makes sense to start applying them at least for the new changes:
> The naming scheme is a faceted dot-separated string, with the delimited terms as dash-separated words.

PR #67, opened 2019-01-22
Pointing to latest acq models

KVupp:
1) Pointing to Latest acq-models. 
2) correcting the latest camel case changes made to piece schema
3) Modifies the receiving_history view to use the Camel case changes

Note:
https://github.com/folio-org/acq-models/pull/77 needs to be merged first for this PR to pass the build


src/test/java/org/folio/rest/impl/ReceivingHistoryTest.java
```diff
@@ -53,27 +54,28 @@ public void testReceivingHistory() {
...
-     
-  
+
+
       logger.info("--- mod-orders-storage pieces test: After receiving_history View created ...");
       verifyViewCollectionAfter(RECEIVING_HISTORY_ENDPOINT, 2);
 
     } catch (Exception e) {
       logger.error("--- mod-orders-storage-test: receiving_history API ERROR: " + e.getMessage(), e);
+      fail();
```
craigmcnally: whoops, how did this slip through the cracks before?  Please add a message to the fail call... depending on how you run the tests, it's sometimes hard to find the error message that caused the failure.

PR #69, opened 2019-01-23
MODORDSTOR-42 unique index for po_number

grabsefx:
## Purpose
In order to guarantee PO number uniqueness across all orders a unique constraint needs to be added for PO number filed in DB
https://issues.folio.org/browse/MODORDSTOR-42

## Approach
Adding unique index in database creation snippet

### DIFF
--- a/src/main/resources/templates/db_scripts/purchase_order_table.sql
+++ b/src/main/resources/templates/db_scripts/purchase_order_table.sql
@@ -1,4 +1,4 @@
 CREATE SEQUENCE IF NOT EXISTS ${myuniversity}_${mymodule}.po_number MAXVALUE 9999999999999999 START WITH 10000 CACHE 1 NO CYCLE;
 GRANT USAGE ON SEQUENCE ${myuniversity}_${mymodule}.po_number TO ${myuniversity}_${mymodule};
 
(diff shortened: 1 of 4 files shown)


src/main/resources/templates/db_scripts/purchase_order_table.sql
```diff
@@ -1,2 +1,4 @@
 CREATE SEQUENCE IF NOT EXISTS ${myuniversity}_${mymodule}.po_number MAXVALUE 9999999999999999 START WITH 10000 CACHE 1 NO CYCLE;
 GRANT USAGE ON SEQUENCE ${myuniversity}_${mymodule}.po_number TO ${myuniversity}_${mymodule};
+
+CREATE UNIQUE INDEX purchase_order_po_number_idx ON ${myuniversity}_${mymodule}.purchase_order ((jsonb->>'po_number') );
```
craigmcnally: Please add `IF NOT EXISTS` to avoid problems during module upgrade

craigmcnally: Also, it seems like the convention is to name unique indexes with the `_idx_unique` or `_unique_idx` suffix

PR #71, opened 2019-01-24
Modordstor 49 refactor sample data from flat files into individual json files

pautau-s:
## Purpose
[MODORDSTOR-49](https://issues.folio.org/browse/MODORDSTOR-49) Refactor sample data from flat files into individual json files.

## Approach
- split flat files into individual json files
- put sample data for each table in a separate folder
- gave readable names to json files with sample data for PO lines and purchase orders

PR #72, opened 2019-01-24
MODORDSTOR-48 - Refactor sample data load

KVupp:
https://issues.folio.org/browse/MODORDSTOR-48

## Purpose
To load sample data via /_/tenant interface.

## Approach
1) Sample data can be loaded using any of the following methods
     i)  parameter in the body to /_/tenant API "parameters": [ { **"key": "loadSample", "value": "true"** }]
     ii) a command line argument via java -jar target/mod-orders-storage-fat.jar **loadSample=true**
     iii) call to /_/proxy/tenants/test-tenant/install?**tenantParameters=loadSample=true** (if enabling via okapi)
2) Modified the unit test case to support an extra header **X-Okapi-Url-to** to support tenant implementation
3) When a module is being upgraded, it will update the sample data, and if not present it will insert it

##Additional changes
1) Removed the old version of Rest assured jar which was duplicate
2) Corrected Receiving history test case which was dependent on the order of response. 

#### TODOS and Open Questions
Will be merged along with https://github.com/folio-org/mod-orders-storage/pull/71, as this has the data to support this pull request

## Learning
https://github.com/folio-org/mod-inventory-storage/pull/215/files This PR was taken as a base implementation


src/main/java/org/folio/rest/impl/TenantReferenceAPI.java
```diff
@@ -0,0 +1,249 @@
+package org.folio.rest.impl;
+
+import static java.util.stream.Collectors.toList;
+import static org.folio.rest.RestVerticle.MODULE_SPECIFIC_ARGS;
```
craigmcnally: I didn't realize this existed, but it seems it's use if fairly common.

README.md
```diff
@@ -9,6 +9,56 @@ Version 2.0. See the file "[LICENSE](LICENSE)" for more information.
 
 This is the Orders storage module.
 
+## PREPARATION
+
+### Tenant Initialization
```
craigmcnally: I suggest you rename this section to Sample Data

README.md
```diff
@@ -9,6 +9,56 @@ Version 2.0. See the file "[LICENSE](LICENSE)" for more information.
...
+
+A tenant can be initialized in the following ways and a parameter can be passed to control the Sample data load
```
craigmcnally: Maybe start by stating that sample data is optionally loaded during tenant initialization... then continue with what you have here...

README.md
```diff
@@ -9,6 +9,56 @@ Version 2.0. See the file "[LICENSE](LICENSE)" for more information.
...
+If no parameters are passed the sample data will not be loaded. 
+
+The module supports v1.2 of the Okapi _tenant interface. This version of the interface allows Okapi to pass tenant initialization parameters using the tenantParameters key. Currently, the only parameter supported is the loadReference key, which will cause the module to load reference data for the tenant if set to true. Here is an example of passing the parameter to the module via Okapi's /_/proxy/tenants/<tenantId>/install endpoint:
```
craigmcnally: loadSample not loadReference.

README.md
```diff
@@ -9,6 +9,56 @@ Version 2.0. See the file "[LICENSE](LICENSE)" for more information.
...
+  "parameters": [
+    {
+      "key": "loadSample",
+      "value": "true"
+    }
+  ]
+}
+```
+
+* COMMAND LINE:
```
craigmcnally: I would split this out into a separate sub-section and explain that it's a way to change the default behavior, possibly cover the entire chain of fallbacks/defaults

src/main/java/org/folio/rest/impl/TenantReferenceAPI.java
```diff
@@ -0,0 +1,249 @@
...
+
+  @Override
+  public void postTenant(TenantAttributes tenantAttributes, Map<String, String> headers, Handler<AsyncResult<Response>> hndlr, Context cntxt) {
+    log.info("postTenant");
+
+    httpClient = cntxt.owner().createHttpClient();
+    super.postTenant(tenantAttributes, headers, res -> {
+      //if a system parameter is passed from command line, ex: loadSample=true that value is considered,
+      //Priority of Parameter Tenant Attributes > command line parameter > default (false)
+      boolean loadSample = Boolean.parseBoolean(MODULE_SPECIFIC_ARGS.getOrDefault(PARAMETER_LOAD_SAMPLE,
```
craigmcnally: There's no need to read/parse the command line arg each time... just do it once

src/main/java/org/folio/rest/impl/TenantReferenceAPI.java
```diff
@@ -0,0 +1,249 @@
...
+          "false"));
+      List<Parameter> parameters = tenantAttributes.getParameters();
+      for (Parameter parameter : parameters) {
+        if (PARAMETER_LOAD_SAMPLE.equals(parameter.getKey())
+          && "true".equals(parameter.getValue())) {
```
craigmcnally: equalsIgnoreCase would add flexibility here

src/main/java/org/folio/rest/impl/TenantReferenceAPI.java
```diff
@@ -0,0 +1,249 @@
...
+        hndlr.handle(io.vertx.core.Future.succeededFuture(PostTenantResponse
+            .respond500WithTextPlain(exception.getLocalizedMessage())));
+      }
+
+  }
+
+  /**
+   * Read the sub directories under "data/" which contains data to be loaded from jar or Path
+   *
+   * @return
```
craigmcnally: returns what?

src/main/java/org/folio/rest/impl/TenantReferenceAPI.java
```diff
@@ -0,0 +1,249 @@
...
+      loadSampleDataForEndpoint(headers, endPoint, asyncResult -> {
+        if (asyncResult.failed()) {
+          res.handle(Future.failedFuture(asyncResult.cause()));
+        } else {
+          loadSampleData(headers, iterator, res);
+        }
+      });
+    }
+  }
+  /**
```
craigmcnally: empty line between method and comment block

src/main/java/org/folio/rest/impl/TenantReferenceAPI.java
craigmcnally: I wonder if there's a reasonable way to test these args in JUnit tests

src/main/java/org/folio/rest/impl/TenantReferenceAPI.java
KVupp: Done

README.md
KVupp: Done

README.md
KVupp: Done

README.md
KVupp: Done

README.md
KVupp: Done

craigmcnally:
@KVupp a problem was raised by @pautau-s at the standup... With this approach we'll re-introduce the problem of trying to insert records that already exist upon upgrade.  I think we would know based on the request whether or not this is an install/enable or an upgrade.  In the latter case, if loadSample is true we might try to delete the sample data records before posting them.

KVupp:
> @KVupp a problem was raised by @pautau-s at the standup... With this approach we'll re-introduce the problem of trying to insert records that already exist upon upgrade. I think we would know based on the request whether or not this is an install/enable or an upgrade. In the latter case, if loadSample is true we might try to delete the sample data records before posting them.

@craigmcnally The code is in place Now, I am doing a PUT when the module is being upgraded and if the data was not present initially it will be Inserted

src/main/java/org/folio/rest/impl/TenantReferenceAPI.java
craigmcnally: right, it's done every time the endpoint is called.  Since this is a command line argument it's not going to change unless the application is restarted, so you can read this in as a static variable...

src/main/java/org/folio/rest/impl/TenantReferenceAPI.java
```diff
@@ -0,0 +1,291 @@
...
+        req.headers().add(header, headerEntry.getValue());
+      }
+    }
+    req.headers().add("Content-Type", "application/json");
+    req.headers().add("Accept", "application/json, text/plain");
+    req.end(json);
+  }
+
+
+  private void putData(Map<String, String> headers, final String endPointUrl, String json, Future<Void> future) {
```
craigmcnally: Not a huge deal but it would be nice to abstract out the duplicate code between putData and postData

src/main/java/org/folio/rest/impl/TenantReferenceAPI.java
KVupp: Ahh right, Done

src/main/java/org/folio/rest/impl/TenantReferenceAPI.java
KVupp: Sure

KVupp:
> Code coverage on new code needs to be higher. Sonarqube is also complaining about a code smell.
> 
> https://sonarcloud.io/project/issues?id=org.folio%3Amod-orders-storage&pullRequest=72&resolved=false

Upon a closer look I understood that the less code coverage is because the data needed for covering all the test cases is present in another PR (https://github.com/folio-org/mod-orders-storage/pull/71), as a result all the code is not being covered. So merging the 2 dependent branches

KVupp:
> X-Okapi-User-id

If the module is being enabled on an entire application it will be going through okapi in which case there will always be a x-okapi-token. 
In cases were the module is run standalone, yes I think the user will have to pass the parameter explicitly.
Also, I have added this parameter in test class where we are creating a tenant

craigmcnally:
As discussed ~75% coverage on new code is close enough for now given that this story is blocking other work.

PR #77, opened 2019-01-31
Modordstor 41- Fix sample data UUID references

KVupp:
https://issues.folio.org/browse/MODORDSTOR-41

## Purpose
To correct the wrong UUIDs present in the sample data
The UUID references are pointing to other records in the sample data
## Approach
The UUID formats were fixed as part of https://github.com/folio-org/mod-orders-storage/pull/72
This story is mainly aimed at corresponding the records with each other in the sample data. 
Also renamed each sub object file to reflect the PO line and PO it maps to, there by making it easier to track all the information


src/main/resources/data/purchase_orders/268758_one-time_open.json
```diff
@@ -5,7 +5,7 @@
   "po_number": "268758",
   "order_type": "One-Time",
   "re_encumber": false,
-  "total_estimated_price": 127.94,
+  "total_estimated_price": 444.95,
```
craigmcnally: Neither of these values match what's in the test data spreadsheet, but I think as long as the cost adds up correctly it should be fine.

craigmcnally: Actually I see this is the for many of the total_estimated_price fields... still probably not a huge deal, just wanted to make a note of it.

KVupp: I never knew the existence of this spreadsheet until now. I just made sure all the total_estimated_price match what's on their corresponding po line files

PR #74, opened 2019-01-29
MODORDSTOR-46 - Create RAML/modDescriptor for po_line_number

khandramai:
## Purpose
According [MODORDSTOR-46](https://issues.folio.org/browse/MODORDSTOR-46) the mod-order-storage RAML and moduleDescriptor should be changed to support implementation of the po_line_number API.

## Approach
- RAML [po_line_number.raml](https://github.com/folio-org/mod-orders-storage/blob/MODORDSTOR-46/ramls/po_line_number.raml) created/updated for this API
- The [sequence_number schema](https://github.com/folio-org/acq-models/commit/99bdb24c61c29908a21ca2026211b9c23986b302) was created with example
- The module descriptor [ModuleDescriptor-template.json](https://github.com/folio-org/mod-orders-storage/blob/MODORDSTOR-46/descriptors/ModuleDescriptor-template.json) updated w/ the new endpoint, permissions, etc.

PR #78, opened 2019-01-31
Modorders 105- Receiving History API

KVupp:
https://issues.folio.org/browse/MODORDERS-105

## Purpose
The receiving history permissions are not being included when "orders-storage.module.all" is added for a tenant. 

## Approach
Correcting the broken receiving history permissions because of duplicate permissions

##TODO
To be merged along with https://github.com/folio-org/mod-orders/pull/94


piotr-kalashuk:
Similar issue was noticed by @khandramai in https://github.com/folio-org/mod-orders-storage/pull/74. Merging this PR as it does not affect other modules (but even makes situation better :) )

PR #99, opened 2019-03-27
MODORDSTOR-68 - Piece records should have format of the piece, not the poLine

khandramai:
[MODORDSTOR-68](https://issues.folio.org/browse/MODORDSTOR-68) - Piece records should have format of the piece, not the poLine

## Purpose
The receiving history has the poLine's order format, but for orders that aren't all physical or all electronic, it becomes ambiguous during receiving which piece you're actually receiving. In order to avoid this confusion we need to make changes to both piece (schema) and receiving history (schema and view).

## Approach
- Unit tests samples for pieces and receiving history updated
- Unit test for recieving history updated
- DB script updated: removed **receivingHistory.poLineOrderFormat**, added **receivingHistory.pieceFormat**

#### TODO and Open Questions
- Merge together with https://github.com/folio-org/mod-orders/pull/131

## Pre-Merge Checklist


src/test/java/org/folio/rest/impl/ReceivingHistoryTest.java
```diff
@@ -137,6 +137,7 @@ private void verifyFields(Piece piece, PoLine poLine, PurchaseOrder purchaseOrde
     assertEquals(poLine.getPurchaseOrderId(), receivingHistory.getPurchaseOrderId());
     assertEquals(purchaseOrder.getDateOrdered(), receivingHistory.getDateOrdered());
     assertEquals(poLine.getOrderFormat().value(), receivingHistory.getPoLineOrderFormat().value());
```
piotr-kalashuk: As per final discussion with UI team, please remove order format as they have to get PO Line for check-in flow anyway

khandramai: Done

PR #102, opened 2019-04-23
MODORDSTOR-22 - Implement GET /orders w/ search and filtering

khandramai:
[MODORDSTOR-22](https://issues.folio.org/browse/MODORDSTOR-22)

## Purpose
In order to support search and filtering across the various orders tables, we'll need to create a view and GET /orders API is for querying that view. 

## Approach
Existing GET /orders API refactored to allow search and filtering.

## Pre-Merge Checklist

  - [ ] Code coverage on new code is 80% or greater
  - [ ] Duplications on new code is 3% or less
  - [ ] There are no major code smells or security issues
  - [ ] Were permissions changed, added, or removed?
  - [ ] There are no breaking changes in this PR.

### DIFF
--- a/src/main/java/org/folio/rest/persist/QueryHolder.java
+++ b/src/main/java/org/folio/rest/persist/QueryHolder.java
@@ -24,6 +24,10 @@ public class QueryHolder {
     this.lang = lang;
   }
 
+  public QueryHolder(String table, String query, int offset, int limit, String lang) {
+    this(table, HelperUtils.JSONB, query, offset, limit, lang);
+  }
+
   public String getTable() {
 
     return table;
--- a/src/main/java/org/folio/rest/persist/QueryHolder.java
+++ b/src/main/java/org/folio/rest/persist/QueryHolder.java
@@ -9,13 +9,15 @@ import org.z3950.zing.cql.cql2pgjson.FieldException;
 public class QueryHolder {
 
   private String table;
+  private String searchField;
   private String query;
   private int offset;
   private int limit;
   private String lang;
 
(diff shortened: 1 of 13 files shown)


craigmcnally:
I've accepted this as everything looks good and coverage is acceptable.  However, I'll add that it would be nice to have coverage on at least some of the exceptional code paths...

khandramai:
@craigmcnally,
you are right - it seems not so easy to cover these exceptional paths of HelperUtils during Orders/Receiving_History unit testing. But I can propose to cover generic static methods from HelperUtils by separate test exactly for this class, probably, with using mock technique. Let me know if we need this now and I can provide some solution in scope of current PR.

src/main/java/org/folio/rest/persist/QueryHolder.java
```diff
@@ -9,13 +9,15 @@
 public class QueryHolder {
 
   private String table;
+  private String searchField;
   private String query;
   private int offset;
   private int limit;
   private String lang;
 
-  public QueryHolder(String table, String query, int offset, int limit, String lang) {
```
piotr-kalashuk: You could leave constructor with default behavior to not modify any other classes:
```java
  public QueryHolder(String table, String query, int offset, int limit, String lang) {
    this(table, HelperUtils.JSONB, query, offset, limit, lang);
  }
```

craigmcnally:
@khandramai I'll let you decide... if you think it's enough work to warrant a separate story, go ahead and create one or let me know and I will.  If you think you can squeeze it into this story w/o much effort even better.

khandramai:
Work on HelperUtils test coverage will be continued in separate PR.

PR #103, opened 2019-04-25
MODORDSTOR-73-Search and filtering for order lines

KVupp:
## Purpose
https://issues.folio.org/browse/MODORDSTOR-73 : Enable search and filtering for /order-lines API

## Approach
create a new view combining purchase order and PO Line table data.

## Pre-Merge Checklist

  - [ ] Were permissions changed, added, or removed?
  
If there are breaking changes, please **STOP** and consider the following:

- What other modules will these changes impact?
- Do JIRAs exist to update the impacted modules?
  - [ ] If not, please create them
  - [ ] Do they contain the appropriate level of detail?  Which endpoints/schemas changed, etc.
  - [ ] Do they have all they appropriate links to blocked/related issues?
- Are the JIRAs under active development?  
  - [ ] If not, contact the project's PO and make sure they're aware of the urgency.
- Do PRs exist for these changes?
  - [ ] If so, have they been approved?

Ideally all of the PRs involved in breaking changes would be merged in the same day to avoid breaking the folio-testing environment.  Communication is paramount if that is to be achieved, especially as the number of intermodule and inter-team dependencies increase.  

While it's helpful for reviewers to help identify potential problems, ensuring that it's safe to merge is ultimately the responsibility of the PR assignee.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,14 @@ java -jar target/mod-orders-storage-fat.jar loadSample=true
 
 Note: The priority for the parameters in the order Tenant Parameters > Command Line argument > Default value
 
-
+### Search and Filtering on APIs
+For the APIs below separate views are created to enable queries that contain fields from Composite Purchase Order.
+* /orders
+   A separate view is created with a metadata column that combines the Purchase order and PO Line JSONs against which queries are executed.
+   It returns the Purchase Order Collection results that match the query
+* /orders-storage/po-lines
+   A separate view is created with a metadata column that combines the Purchase order and PO Line JSONs against which queries are executed.
+   It returns the PO Line Collection results that match the query
 
 ## Additional information
 
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ For the APIs below separate views are created to enable queries that contain fie
 * /orders
    A separate view is created with a metadata column that combines the Purchase order and PO Line JSONs against which queries are executed.
    It returns the Purchase Order Collection results that match the query
-* /orders-storage/po-lines
+* /orders-storage/order-lines
    A separate view is created with a metadata column that combines the Purchase order and PO Line JSONs against which queries are executed.
    It returns the PO Line Collection results that match the query
 
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ Note: The priority for the parameters in the order Tenant Parameters > Command L
 
 ### Search and Filtering on APIs
 For the APIs below separate views are created to enable queries that contain fields from Composite Purchase Order.
-* /orders
+* /orders-storage/orders
    A separate view is created with a metadata column that combines the Purchase order and PO Line JSONs against which queries are executed.
    It returns the Purchase Order Collection results that match the query
 * /orders-storage/order-lines
--- a/src/main/java/org/folio/rest/impl/PoLinesAPI.java
+++ b/src/main/java/org/folio/rest/impl/PoLinesAPI.java
@@ -15,11 +15,12 @@ import org.folio.rest.persist.QueryHolder;
 
 import javax.ws.rs.core.Response;
 import java.util.Map;
-
+import static org.folio.rest.persist.HelperUtils.METADATA;
 import static org.folio.rest.persist.HelperUtils.getEntitiesCollection;
 
 public class PoLinesAPI implements OrdersStoragePoLines {
   private static final String POLINE_TABLE = "po_line";
+  private static final String POLINE_VIEW = "order_lines_view";
   private String idFieldName = "id";
 
 
@@ -33,7 +34,7 @@ public class PoLinesAPI implements OrdersStoragePoLines {
       Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
     vertxContext.runOnContext((Void v) -> {
       EntitiesMetadataHolder<PoLine, PoLineCollection> entitiesMetadataHolder = new EntitiesMetadataHolder<>(PoLine.class, PoLineCollection.class, GetOrdersStoragePoLinesResponse.class);
-      QueryHolder cql = new QueryHolder(POLINE_TABLE, query, offset, limit, lang);
+      QueryHolder cql = new QueryHolder(POLINE_VIEW, METADATA, query, offset, limit, lang);
       getEntitiesCollection(entitiesMetadataHolder, cql, asyncResultHandler, vertxContext, okapiHeaders);
     });
   }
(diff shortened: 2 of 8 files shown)


README.md
```diff
@@ -62,7 +62,14 @@ java -jar target/mod-orders-storage-fat.jar loadSample=true
 
 Note: The priority for the parameters in the order Tenant Parameters > Command Line argument > Default value
 
-
+### Search and Filtering on APIs
+For the APIs below separate views are created to enable queries that contain fields from Composite Purchase Order.
+* /orders
```
piotr-kalashuk: Just looking at this change I realized that this endpoint has to be changed to `/orders-storage/orders` to align with [Acquisitions API Listing](https://docs.google.com/spreadsheets/d/1se-a2owRfHLG5eaAZglx4vLvRClCKGp2wZCfD39ZJRY/view#gid=894044560)
@khandramai please prepare PR

src/main/java/org/folio/rest/impl/PoLinesAPI.java
```diff
@@ -33,7 +34,7 @@ public void getOrdersStoragePoLines(String query, int offset, int limit, String
       Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
     vertxContext.runOnContext((Void v) -> {
       EntitiesMetadataHolder<PoLine, PoLineCollection> entitiesMetadataHolder = new EntitiesMetadataHolder<>(PoLine.class, PoLineCollection.class, GetOrdersStoragePoLinesResponse.class);
```
piotr-kalashuk: According to [Acquisitions API Listing](https://docs.google.com/spreadsheets/d/1se-a2owRfHLG5eaAZglx4vLvRClCKGp2wZCfD39ZJRY/view#gid=894044560&range=B48) the `/orders-storage/order-lines` endpoint is going to use view

src/test/java/org/folio/rest/impl/SearchOrderLinesTest.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.List;
+import org.folio.rest.jaxrs.model.Physical;
+import org.folio.rest.jaxrs.model.PoLine;
+import org.folio.rest.jaxrs.model.PoLine.OrderFormat;
+import org.folio.rest.jaxrs.model.PoLineCollection;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+public class SearchOrderLinesTest extends TestBase {
```
piotr-kalashuk: Please add this test class to `StorageTestSuite` because at the moment it is ignored (please have a look at [MODORDSTOR-73 #2](https://jenkins-aws.indexdata.com/job/folio-org/job/mod-orders-storage/job/MODORDSTOR-73/2/testReport/org.folio.rest.impl/)

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -4,11 +4,16 @@
...
+    },
+    {
       "run": "after",
       "snippet": "CREATE OR REPLACE VIEW orders_view AS SELECT purchase_order.id AS id, purchase_order.jsonb AS jsonb, po_line.jsonb || COALESCE(purchase_order.jsonb, '{}'::jsonb) AS metadata FROM po_line LEFT JOIN purchase_order ON (po_line.jsonb -> 'purchaseOrderId'::text) = (purchase_order.jsonb -> 'id'::text);",
       "fromModuleVersion": "1.0"
+    },
+    {
+      "run": "after",
+      "snippet": "CREATE OR REPLACE VIEW order_lines_view AS SELECT po_line.id AS id, po_line.jsonb AS jsonb, po_line.jsonb || COALESCE(purchase_order.jsonb, '{}'::jsonb) AS metadata FROM po_line LEFT JOIN purchase_order ON (po_line.jsonb -> 'purchaseOrderId'::text) = (purchase_order.jsonb -> 'id'::text);",
+      "fromModuleVersion": "5.0"
```
piotr-kalashuk: When reviewing #102 I was going to ask Slava to change `fromModuleVersion` for `orders_view` from `1.0` to `5.0` as well. However when I checked version in latest testing vagrant box I was a bit confuced and decided to leave it as it is. However I still think that proper version should be really `5.0` for any new changes...
image

src/main/java/org/folio/rest/impl/PoLinesAPI.java
KVupp: oh right Thanks, I missed that part

src/test/java/org/folio/rest/impl/SearchOrderLinesTest.java
KVupp: Done

src/main/resources/templates/db_scripts/schema.json
KVupp: That is true but I think RMB decides if it is an upgrade based on the input provided to tenant API, atleast from this line 
https://github.com/folio-org/raml-module-builder/blob/master/domain-models-runtime/src/main/java/org/folio/rest/impl/TenantAPI.java#L244

So even I prefer the proper version to be maintained in the files so that the views are not created atleast in case of older versions.

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -4,11 +4,16 @@
...
-  ,{
+    },
+    {
       "run": "after",
       "snippet": "CREATE OR REPLACE VIEW orders_view AS SELECT purchase_order.id AS id, purchase_order.jsonb AS jsonb, po_line.jsonb || COALESCE(purchase_order.jsonb, '{}'::jsonb) AS metadata FROM po_line LEFT JOIN purchase_order ON (po_line.jsonb -> 'purchaseOrderId'::text) = (purchase_order.jsonb -> 'id'::text);",
       "fromModuleVersion": "1.0"
+    },
+    {
+      "run": "after",
+      "snippet": "CREATE OR REPLACE VIEW order_lines_view AS SELECT po_line.id AS id, po_line.jsonb AS jsonb, po_line.jsonb || COALESCE(purchase_order.jsonb, '{}'::jsonb) AS metadata FROM po_line LEFT JOIN purchase_order ON (po_line.jsonb -> 'purchaseOrderId'::text) = (purchase_order.jsonb -> 'id'::text);",
```
piotr-kalashuk: `po_line.jsonb || COALESCE(purchase_order.jsonb, '{}'::jsonb) AS metadata` will join all properties from both tables. However there is one property `id` which presents in both. I believe that the `id` will be chosen from purchase_order.jsonb.
In case CQL `query` contains `...id=123 or ...` could you please check what SQL will be generated for such query?

KVupp: That is true common key fields will be replaced with the value from the right hand operand. hence metadata.id=purchase_order.id. 
I think it makes sense to change it to be metadata.id=po_line.id

PR #131, opened 2019-06-20
MODORDSTOR-85 - DB Optimization

khandramai:
[MODORDSTOR-85](https://issues.folio.org/browse/MODORDSTOR-85) - DB Optimization

## Purpose
Define index, ginIndex and uniqueIndex, remove generateId property.

## Approach

**fullTextIndexes**

>    fullTextIndex - create a full text index using the tsvector features of postgres. These do their
>    own normalizing, so there is no need to use caseSensitive or removeAccents. The tOps
>    is optional (like for all indexes), and defaults to ADDing the index. whereClause and
>    stringType work as for likeIndex above.

were used for fields with text content:
purchase_order -> closeReason.reason, closeReason.note
po_line -> publisher, donor, requester, selector
piece -> comment, caption

**uniqueIndexes**

> uniqueIndex - create a unique index on a field in the json
>   - the tOps indicates the table operation - ADD means to create this index, DELETE indicates this index should be removed
>    - the whereClause allows you to create partial indexes, for example: "whereClause": "WHERE (jsonb->>'enabled')::boolean = true"
>    - See additional options in the likeIndex section above

were used for unique fields:
po_line -> poLineNumber
purchase_order -> unique index has already defined

**ginIndexes** (with "caseSensitive": true, "removeAccents": false)

> ginIndex - generate an inverted index on the json using the gin_trgm_ops extension. Allows for regex 
> queries to run in an optimal manner (similar to a simple search engine). Note that the generated index > is large and does not support the equality operator (=). See the likeIndex for available options (does not support partial indexes - where). removeAccents is set to true and is not case sensitive.

ginIndexes were used in cases when case sensitivity and accents are actual and for arrays:
po_line -> fundDistribution.code, details.productIds, title, vendor.vendorAccount, vendor.refNumber, tags, contributors 
purchase_order -> unique index has already defined
piece -> caption

**indexes** (with "caseSensitive": false)

>  index - create a btree index on a field in the json
    the tOps indicates the table operation - ADD means to create this index, DELETE indicates this index should be removed
    the whereClause allows you to create partial indexes, for example: "whereClause": "WHERE (jsonb->>'enabled')::boolean = true"
    See additional options in the likeIndex section above

were used for all other cases:
po_line -> eresource.accessProvider,  eresource.activated ,  eresource.expectedActivation ,  eresource.trial , eresource.materialType , details.subscriptionFrom , details.subscriptionTo , 
location.locationId , physical.receiptDue , physical.expectedReceiptDate , physical.volumes , physical.materialType , claim.grace , claim.sent , collection , receiptStatus , paymentStatus , orderFormat ,
receiptDate ,  metadata.createdDate ,  rush ,  source 
purchase_order -> assignedTo, metadata.createdDate, metadata.createdByUserId, workflowStatus, approved, dateOrdered, orderType, reEncumber, renewal.manualRenewal, renewal.renewalDate, renewal.reviewPeriod
piece -> format, itemId, locationId, poLineId, receivingStatus, receivedDate, supplement

Notes:

> the caseSensitive allows you to create case insensitive indexes (boolean true / false), if you have a string field that may have different casings and you want the value to be unique no matter the case. Default: false should not be changed, temporarily all indexes are created with a lower() function wrapper. Index will not be used if this is changed (important on large tables)

Performance testing results:

It was measured 1000 iterations on local environment for initial 10000 po and po_line records for both direct endpoint purchase-orders/po-lines and corresponding view orders/order-lines.

Searching:
1. /orders-storage/purchase-orders?query=poNumber==10073 (/orders-storage/orders?query=poNumber==10073 )
2. /orders-storage/po-lines?query=poLineNumber==10073-1 (/orders-storage/order-lines?query=poLineNumber==10073-1)

Filtering:
1. /orders-storage/purchase-orders?query=workflowStatus==Open (/orders-storage/orders?query=workflowStatus==Open)
2. /orders-storage/po-lines?query=orderFormat=Physical Resource (/orders-storage/order-lines?query=orderFormat=Physical Resource)

Sorting:
1. /orders-storage/orders?query=workflowStatus==Open AND orderFormat=Physical Resource sortBy metadata.createdDate/sort.ascending
2. /orders-storage/order-lines?query=workflowStatus==Open AND orderFormat=Physical Resource sortBy metadata.createdDate/sort.ascending

Label | Average non-indexed, ms | 95% Line non-indexed, ms | Throughput non-indexed | Average indexed, ms | 95% Line indexed, ms | Throughput indexed | Decreasing average, % | Decreasing 95% Line, % | Increasing throughput, %
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
Search : Orders (purchase-orders) | 34 | 45 | 0,21022 | 33 | 46 | 0,21256 | 2,9 | -2,2 | 1,1
Search : Orders (orders) | 935 | 985 | 0,21019 | 918 | 969 | 0,21253 | 1,8 | 1,6 | 1,1
Search : Order-Lines (po-lines) | 165 | 184 | 0,21023 | 168 | 187 | 0,21258 | -1,8 | -1,6 | 1,1
Search : Order-Lines (order-lines) | 569 | 601 | 0,21022 | 567 | 598 | 0,21257 | 0,35 | 0,5 | 1,1
Filtering : Orders (purchase-orders) | 31 | 42 | 0,21025 | 24 | 35 | 0,2126 | 22,6 | 16,7 | 1,1
Filtering : Orders (orders) | 702 | 744 | 0,21022 | 704 | 746 | 0,21257 | -0,3 | -0,3 | 1,1
Filtering : Order-Lines (po-lines) | 162 | 157 | 0,21025 | 137 | 157 | 0,2 | 15,4 | 0 | 1,1
Filtering : Order-Lines (order-lines) | 386 | 403 | 0,21024 | 384 | 400 | 0,2 | 0,5 | 0,7 | 1,1
Sorting : Orders | 931 | 987 | 0,21022 | 928 | 980 | 0,21257 | 0,3 | 0,7 | 1,1
Sorting : Order-Lines | 840 | 888 | 0,21023 | 839 | 882 | 0,2 | 0,1 | 0,7 | 1,1

## Pre-Merge Checklist


src/main/resources/templates/db_scripts/schema.json
```diff
@@ -19,29 +19,79 @@
   "tables": [
     {
       "tableName": "alert",
-      "generateId": true,
```
KVupp: please upgrade the pom.xml to use latest RMB version for this change to take effect

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -19,29 +19,79 @@
...
       "fromModuleVersion": 1.0,
       "withMetadata": true,
       "populateJsonWithId": true,
-      "customSnippetPath": "purchase_order_table.sql"
+      "customSnippetPath": "purchase_order_table.sql",
+      "index": [
+        {
+          "fieldName": "assignedTo",
+          "caseSensitive": false
```
KVupp: optional: The caseSensitive is already false by default, removing it makes it cleaner

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -50,11 +100,156 @@
...
+          "fieldName": "donor"
+        },
+        {
+          "fieldName": "requester"
+        },
+        {
+          "fieldName": "selector"
+        }
+      ],
+      "ginIndex": [
```
KVupp: ginIndexes are used for regex queries, from the UI we would be searching for poLineNumber and requestor, would be better to add those too

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -19,29 +19,79 @@
...
+        }
+      ],
+      "fullTextIndex": [
+        {
+          "fieldName": "closeReason.reason"
+        },
+        {
+          "fieldName": "closeReason.note"
+        }
+      ]
```
KVupp: would also make the UI searched better if a ginIndex is created on poNumber, where wildcard searches are used

src/main/resources/templates/db_scripts/schema.json
craigmcnally: IMO making this explicit makes it clear w/o having any knowledge of the defaults, but I'm fine with it either way.

src/main/resources/templates/db_scripts/schema.json
khandramai: I'm not sure that we need this for poNumber. I have discussed with @aliaksei-chumakou this matter - UI doesn't supports wildcard search on poNumber at the moment. Moreover, unique index already created for poNumber in the schema. So, could you please confirm that this makes sense now?

aliaksei-chumakou: @KVupp we use strict searches from ui on po and po line numbers. So it's more a question to POs if they need wildcard searches.

KVupp: Screen Shot 2019-06-21 at 7 39 31 AM

I see a query like "*0" working.
 Apart from all that, any query in the search bar forms a query like the below:
Screen Shot 2019-06-21 at 7 41 55 AM

I will have to look at how CQL converts these queries to SQL queries, but I would assume they would be regex queries

aliaksei-chumakou: @KVupp We don't support such search query from UI. It's a security issue, as for me, when user input isn't escaped and goes directly to database query

aliaksei-chumakou: https://xkcd.com/327/

KVupp: Also from [RMB](https://github.com/folio-org/raml-module-builder#important-information)
> Behind the scenes, the CQL to Postgres query converter will generate regex queries for = >queries. For example: ?query=fieldA=ABC will generate an SQL regex query, which will require >a gin index to perform on large tables.

so from the CQL queries from the UI, it looks like we are always using "=" , which would generate regex queries on SQL, so I think it would still makes sense to create ginIndexes

aliaksei-chumakou: @KVupp Does it make sense to use `=` on UI for strict equal? Maybe it's better to fix UI to use `==` ?

KVupp: @aliaksei-chumakou ,Like you said I would prefer the POs answer this question, because it comes down to what kind of searches we want to support. 
Atleast on inventory app, these same searches are supported and queries are formed in the similar manner

craigmcnally: We should definitely get some input from the POs, but my guess is that users are going to want/expect partial matches w/o explicitly using wildcards; at least for some searches...  so if they type in `52` in the search box for a po number search, they'd get results like 10052, 10520, 15200, etc.  

I would have thought that the UI would achieve this by adding (escaped if needed) wildcards to both ends of the search criteria, and using an `=`; e.g. `poNumber=\*52\*`

craigmcnally: hmm doesn't look like you need to escape the asterisks...

```$ curl $OKAPI/orders/composite-orders --data-urlencode 'query=poNumber=*17*' -H "X-Okapi-Token: $TOKEN" -w '\n' -G
{
  "purchaseOrders" : [ {
    "id" : "4d674341-c40b-4e1c-a993-9d6e8372abd3",
    "approved" : false,
    "dateOrdered" : "2019-06-21T13:06:02.762+0000",
    "notes" : [ "Open Order for Negative API Tests" ],
    "poNumber" : "10170",
    "orderType" : "One-Time",
    "reEncumber" : false,
    "vendor" : "197c79c2-993c-4130-8ae3-a3a8899794d3",
    "workflowStatus" : "Open",
    "metadata" : {
      "createdDate" : "2019-06-21T13:06:02.741+0000",
      "createdByUserId" : "00000001-1111-5555-9999-999999999999",
      "updatedDate" : "2019-06-21T13:06:02.792+0000",
      "updatedByUserId" : "00000001-1111-5555-9999-999999999999"
    }
  }, {
    "id" : "27b3359d-e92e-43f5-b5ff-0fa46bd0bb3a",
    "approved" : false,
    "notes" : [ "Closed Order for Negative API Tests" ],
    "poNumber" : "10171",
    "orderType" : "One-Time",
    "reEncumber" : false,
    "vendor" : "197c79c2-993c-4130-8ae3-a3a8899794d3",
    "workflowStatus" : "Closed",
    "metadata" : {
      "createdDate" : "2019-06-21T13:06:05.585+0000",
      "createdByUserId" : "00000001-1111-5555-9999-999999999999",
      "updatedDate" : "2019-06-21T13:06:05.585+0000",
      "updatedByUserId" : "00000001-1111-5555-9999-999999999999"
    }
  }, {
    "id" : "a727f69c-6e40-40fe-906e-6ea64df78f42",
    "approved" : false,
    "notes" : [ "Pending Order for Negative API Tests" ],
    "poNumber" : "10172",
    "orderType" : "One-Time",
    "reEncumber" : false,
    "vendor" : "197c79c2-993c-4130-8ae3-a3a8899794d3",
    "workflowStatus" : "Pending",
    "metadata" : {
      "createdDate" : "2019-06-21T13:06:07.975+0000",
      "createdByUserId" : "00000001-1111-5555-9999-999999999999",
      "updatedDate" : "2019-06-21T13:06:07.975+0000",
      "updatedByUserId" : "00000001-1111-5555-9999-999999999999"
    }
  }, {
    "id" : "cdb6d204-e12d-43e3-b7dd-fd4acf0f89fc",
    "approved" : true,
    "assignedTo" : "ab18897b-0e40-4f31-896b-9c9adc979a88",
    "notes" : [ "ABCDEFGHIJKLMNO", "ABCDEFGHIJKLMNOPQRST", "ABCDEFGHIJKLMNOPQRSTUV" ],
    "poNumber" : "10174",
    "orderType" : "One-Time",
    "reEncumber" : false,
    "vendor" : "8bb4efff-e6db-446e-8708-4cd14e5625dc",
    "workflowStatus" : "Pending",
    "metadata" : {
      "createdDate" : "2019-06-21T13:08:48.391+0000",
      "createdByUserId" : "00000001-1111-5555-9999-999999999999",
      "updatedDate" : "2019-06-21T13:08:48.391+0000",
      "updatedByUserId" : "00000001-1111-5555-9999-999999999999"
    }
  } ],
  "totalRecords" : 4
}
```

khandramai: OK, 
so let it be ginIndex for poNumber/poLineNumber at least for now.
Thanks for clarification!

khandramai:
@craigmcnally 
I have documented details of this PR on separate wiki page https://wiki.folio.org/display/FOLIJET/Investigation%3A+mod-orders-storage+performance.

src/main/resources/templates/db_scripts/schema.json
craigmcnally: Leave as-is for now.  If it turns out that A) these searches are required and B) are too slow, we can revisit this.

craigmcnally: I suspect that the (unique) btree index will provide sufficient performance, even with these types of queries.  If supported, a ginindex would likely provide even greater performance, but postgres apparently doesn't support unique ginindex (thanks for the link @khandramai).

PR #154, opened 2019-09-09
MODORDSTOR-104 - Release 8.0.0

khandramai:
MODORDSTOR-104 - Release 8.0.0
NB: Major version increased due to Assignments API was deleted, no interface changing just introduced new Order Template API with interface version 1.0. Log4J configuration was changed in accordance with https://github.com/folio-org/mod-invoice-storage/pull/42/commits/492e621c7bb7121a4709348d1d2868b5a183db55.


src/main/resources/log4j.properties
```diff
@@ -1,8 +1,15 @@
-# variables are substituted from filters-[production|development].properties
-log4j.rootLogger=INFO, CONSOLE
-#log4j.rootLogger=DEBUG, CONSOLE
+status = info
```
KVupp: I think this file has to be named log4j**2**.properties as we are using log4j2 logger. Not sure if log4j works too

NEWS.md
```diff
@@ -1,4 +1,18 @@
-## 8.0.0 - Unreleased
+## 9.0.0 - Unreleased
+
+## 8.0.0 - Released
+The main focus of this release was to implement API for order templates and to clean-up API for acquisition-unit-assignments.
+
+[Full Changelog](https://github.com/folio-org/mod-orders-storage/compare/v7.0.0...v8.0.0)
+
+### Stories
+* [MODORDSTOR-106](https://issues.folio.org/browse/MODORDSTOR-106) - Create order-templates API
```
KVupp: https://github.com/folio-org/mod-orders-storage/blob/00db9642f576efefdce5bdf696256a9123f2a53b/src/main/resources/templates/db_scripts/schema.json#L363
Also please change this to 8

src/main/resources/log4j.properties
khandramai: Fixed.

NEWS.md
khandramai: Updated.

khandramai:
@piotr-kalashuk, Regarding to acquisitions-units-storage.units the contract formally remains the same. Even if we do not explicitly specify field isDeleted then it is false by default. Are you OK with interface version 1.1?

PR #249, opened 2021-05-27
MODORDSTOR-231: Upgrade test (fund code migration) with Okapi path

julianladisch:
Fix null body handling in RestClient.get.

Add UpgradeTenantTest.java that runs an upgrade using POST `/_/tenant` API and with an Okapi address with path.


nixorg (approved):

AndreiMPV:
@julianladisch 
RestClient this is wrapper for HttpClient and removes boilerplate logic with other module invocation by REST.
RestClient must return responses as is from HttpClient  and "null" is legal for us.
We will improve our logging and change "completeExceptionally" block with logic provided by you.
Thanks for your effort

PR #251, opened 2021-06-11
MODORDSTOR-232: receiving_history_view should use id primary key index

julianladisch:
Replacing `jsonb->>'id'` by `id` uses the primary key index and is much faster.


julianladisch:
Thunderjet has not time to review external pull request.

PR #313, opened 2022-08-03
[MODORDSTOR-309] Already exported orders are included repeatedly in next exports

SerhiiNosko:
## Purpose
https://issues.folio.org/browse/MODORDSTOR-309

## Approach
There were 3 separate issues fixed to address MODORDSTOR-309
1. Kafka host/port were not configured on https://bulk-edit-perf.ci.folio.org/ for module mod-orders-storage. After add configuration we see that they fetched from secrets as for other modules:
image
2. x-okapi-tenant has not been populated  in all cases until issues was fixed in scope of PR:
https://github.com/folio-org/mod-data-export-spring/pull/203
3. x-okapi-tenant header comes in lower case, but in order-storage its declared in mixed case like X-Okapi-Tenant. Here is list of headers coming in kafka record:
image
This particular PR fixes issue this third issue

### DIFF
--- a/src/main/java/org/folio/event/KafkaEventUtil.java
+++ b/src/main/java/org/folio/event/KafkaEventUtil.java
@@ -12,7 +12,7 @@ public final class KafkaEventUtil {
 
   public static String extractValueFromHeaders(List<KafkaHeader> headers, String key) {
     return headers.stream()
-      .filter(header -> header.key().equals(key))
+      .filter(header -> header.key().equalsIgnoreCase(key))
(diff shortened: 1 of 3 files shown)


okolawole-ebsco (approved):

okolawole-ebsco:
is a test possible?

SerhiiNosko:
> is a test possible?

I've updated existing test to cover case from this issue, thanks

SinghAdes (approved):

PR #331, opened 2023-01-06
[MODORDSTOR-325][MODORDSTOR-326][MODORDSTOR-332] Replace PGUtil for Edit Order/Order Line to use PostgresClient directly / Implement lock mechanism for Outbox Event Log

SerhiiNosko:
https://issues.folio.org/browse/MODORDSTOR-325
https://issues.folio.org/browse/MODORDSTOR-326
https://issues.folio.org/browse/MODORDSTOR-332
Major points/changes in this PR:

- Decreased timer interface invocation to 30 mins to not overload DB invoking it each 2 sec
- Instead of added call to process outbox table in onComplete hook after transaction commited
- Refactor approach to work with transaction to avoid deprecate RMB methods with managing transactions manually
- Add lock mechanism for processing events from outbox table using SELECT FOR UPDATE construction


Abdulkhakimov (approved):

src/main/java/org/folio/event/service/AuditOutboxService.java
```diff
@@ -48,87 +53,70 @@ public AuditOutboxService(AuditOutboxEventsLogRepository repository,
...
    * @return future with saved outbox log in the same transaction
    */
-  public Future<Tx<PurchaseOrder>> saveOrderOutboxLog(Tx<PurchaseOrder> tx, OrderAuditEvent.Action action, Map<String, String> okapiHeaders) {
-    Promise<Tx<PurchaseOrder>> promise = Promise.promise();
-    String order = Json.encode(tx.getEntity());
-    saveOutboxLog(tx.getConnection(), action.value(), EntityType.ORDER, order, okapiHeaders)
+  public Future<Void> saveOrderOutboxLog(Conn conn, PurchaseOrder entity, OrderAuditEvent.Action action, Map<String, String> okapiHeaders) {
+    Promise<Void> promise = Promise.promise();
+    String order = Json.encode(entity);
+    saveOutboxLog(conn, action.value(), EntityType.ORDER, order, okapiHeaders)
```
grabsefx, line 96: Please use return statement instead of promise here, or handle promise failure

SinghAdes (approved):

src/main/java/org/folio/services/lines/PoLinesService.java
```diff
@@ -400,25 +462,27 @@ private void populateTitleBasedOnPackagePoLine(Title title, PoLine packagePoLine
...
+  private Future<PoLine> updateTitle(Conn conn, PoLine poLine) {
     Promise<Results<Title>> promise = Promise.promise();
-    PoLine poLine = poLineTx.getEntity();
     Criterion criterion = getCriteriaByFieldNameAndValueNotJsonb(POLINE_ID_FIELD, poLine.getId());
 
-    client.getPgClient().get(poLineTx.getConnection(), TITLES_TABLE, Title.class, criterion, true, false, promise);
+    conn.get(TITLES_TABLE, Title.class, criterion, true)
+      .onComplete(promise);
+
     return promise.future()
```
grabsefx, line 472: it looks like promise variable is not needed anymore. You could just use 
`conn.get(TITLES_TABLE, Title.class, criterion, true).compose(...)
`
Line `.recover(Future::failedFuture);` does nothing and can be removed

SerhiiNosko: Fixed, thank you

grabsefx (approved):

PR #380, opened 2024-01-30
[MODORDSTOR-373] - Add validation for custom field values

alb3rtino:
## Purpose
Values for custom fields need to be valid.
See https://issues.folio.org/browse/MODORDSTOR-373.

## Approach
The `POST` and `PUT` API methods for entities that may include custom field data have been revised. Prior to saving or updating entities, custom field values undergo validation. The validation process utilizes [`ValidationServiceImpl`](https://github.com/folio-org/folio-custom-fields/blob/master/src/main/java/org/folio/validate/ValidationServiceImpl.java) from `folio-custom-fields`.

## Pre-Merge Checklist:


src/main/java/org/folio/rest/impl/OrderTemplatesAPI.java
```diff
@@ -28,7 +31,17 @@ public void getOrdersStorageOrderTemplates(String query, String totalRecords, in
   @Override
   @Validate
   public void postOrdersStorageOrderTemplates(OrderTemplate entity, Map<String, String> okapiHeaders, Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
-    PgUtil.post(ORDER_TEMPLATES_TABLE, entity, okapiHeaders, vertxContext, PostOrdersStorageOrderTemplatesResponse.class, asyncResultHandler);
+    validateCustomFields(vertxContext, okapiHeaders, entity)
+        .compose(
+            v ->
```
SerhiiNosko, line 36: please use the same code style as in other places, many people contributing to this module, and we should have some standard, to keep it we need to follow already written code

SerhiiNosko:
@alb3rtino please attach screens of karate tests execution of this branch if it possible

src/test/java/org/folio/rest/impl/EntititesCustomFieldsTest.java
```diff
@@ -0,0 +1,302 @@
...
+            PURCHASE_ORDER.getEndpoint(),
+            PURCHASE_ORDER.getSampleEntity().encode(),
+            headers)
+        .then()
+        .statusCode(201);
+  }
+
+  void setupOrderLine() throws MalformedURLException {
+    postData(PO_LINE.getEndpoint(), PO_LINE.getSampleEntity().encode(), headers)
+        .then()
```
SerhiiNosko, line 53: in module we use 2 spaces instead of 4 for tabulation

src/main/java/org/folio/rest/core/BaseApi.java
```diff
@@ -33,18 +42,45 @@ public Future<Response> buildErrorResponse(Throwable throwable) {
...
+      .getMap()
+      .entrySet()
+      .stream()
+      .collect(Collectors.toMap(Entry::getKey, Entry::getValue));
+  }
+
+  public Future<Void> validateCustomFields(
+    Context vertxContext, Map<String, String> okapiHeaders, Object entity) {
+    Map<String, Object> customFieldsMap = getCustomFields(entity);
+    if (!customFieldsMap.isEmpty()) {
```
SerhiiNosko, line 78: we can add info log here that we started validation of custom fields, it can help match exception that it relies with custom fields validaiton, and please check that exception with invalid custom field logged

src/main/java/org/folio/rest/impl/OrderTemplatesAPI.java
alb3rtino: I understand the importance of thorough logging, but I'm not entirely convinced that adding an additional log message would provide a significant benefit.

The exception is named `CustomFieldValidationException`, indicating an error related to custom fields and validation. It is logged by the calling API methods, for example [here](https://github.com/folio-org/mod-orders-storage/pull/380/files#diff-d16050af5f2ff623e9f6459f8f6e3d126a10421ba6202067162d5e2032e061d7R75-R77). The response generated by `buildErrorResponse` provides detailed information about the validation error(s).

Considering the provided context and the existing logging mechanisms, do you still think it's necessary to include the additional log message?

src/main/java/org/folio/rest/core/BaseApi.java
alb3rtino: I understand the importance of thorough logging, but I'm not entirely convinced that adding an additional log message would provide a significant benefit.

The exception is named `CustomFieldValidationException`, indicating an error related to custom fields and validation. It is logged by the calling API methods, for example [here](https://github.com/folio-org/mod-orders-storage/pull/380/files#diff-d16050af5f2ff623e9f6459f8f6e3d126a10421ba6202067162d5e2032e061d7R75-R77). The response generated by `buildErrorResponse` provides detailed information about the validation error(s).

Considering the provided context and the existing logging mechanisms, do you still think it's necessary to include the additional log message?

alb3rtino:
> @alb3rtino please attach screens of karate tests execution of this branch if it possible

Unfortunately, I don't have the resources to run a fully-fledged local Vagrant box.

src/main/java/org/folio/rest/core/BaseApi.java
alb3rtino: I've updated the logging in `OrderTemplatesAPI` to align with the rest of the APIs.

src/main/java/org/folio/rest/core/BaseApi.java
```diff
@@ -4,16 +4,25 @@
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
 import static javax.ws.rs.core.MediaType.TEXT_PLAIN;
 import static javax.ws.rs.core.Response.Status.INTERNAL_SERVER_ERROR;
+import static org.folio.rest.tools.utils.TenantTool.tenantId;
```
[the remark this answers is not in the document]
alb3rtino: Done. I believe it could be beneficial for both maintainers and contributors if you could point to any specific documents, guides, or configuration files that outline the preferred formatting and conventions you are using. Currently, `CONTRIBUTING.md` only links to the generic FOLIO guidelines.

imerabishvili (approved):

PR #384, opened 2024-03-06
[MODORDSTOR-379] - Manage updates of custom fields through RecordServiceImpl

alb3rtino:
## Purpose
https://folio-org.atlassian.net/browse/MODORDSTOR-379

This pull request switches the implementation used for `RecordService` from the default `NoOpRecordService` to `RecordServiceImpl`. See the [README](https://github.com/folio-org/folio-custom-fields?tab=readme-ov-file#recordservice) section in [folio-custom-fields](https://github.com/folio-org/folio-custom-fields).

`RecordServiceImpl` ensures that if a user deletes or updates a custom field, these changes will be appropriately reflected in the entities that have values set for this field. Basically, if a custom field is removed it also gets removed from its entities. If a selectable option is removed from a custom field then its entities will no longer have that option selected. 

## Approach
- Implemented `RecordServiceFactory` to return an instance of `RecordServiceImpl`.
- Added tests to validate the functionality of the implementation.

## Pre-Merge Checklist:

If there are breaking changes, please **STOP** and consider the following:

- What other modules will these changes impact?
- Do JIRAs exist to update the impacted modules?
  - [ ] If not, please create them
  - [ ] Do they contain the appropriate level of detail?  Which endpoints/schemas changed, etc.
  - [ ] Do they have all the appropriate links to blocked/related issues?
- Are the JIRAs under active development?
  - [ ] If not, contact the project's PO and make sure they're aware of the urgency.
- Do PRs exist for these changes?
  - [ ] If so, have they been approved?
- Did you modify code to call some additional endpoints?
  - [ ] If so, do you check that necessary module permission added in ModuleDescriptor-template.yaml file?

Ideally, all the PRs involved in breaking changes would be merged on the same day
to avoid breaking the folio-testing environment.
Communication is paramount if that is to be achieved,
especially as the number of inter-module and inter-team dependencies increase.

While it's helpful for reviewers to help identify potential problems,
ensuring that it's safe to merge is ultimately the responsibility of the PR assignee.

### DIFF
--- a/src/main/java/org/folio/services/record/RecordServiceFactoryImpl.java
+++ b/src/main/java/org/folio/services/record/RecordServiceFactoryImpl.java
@@ -1,0 +1,24 @@
+package org.folio.services.record;
+
+import java.util.List;
+import java.util.Map;
+
+import org.folio.service.RecordService;
+import org.folio.service.RecordServiceImpl;
+import org.folio.service.spi.RecordServiceFactory;
+
+import io.vertx.core.Vertx;
+
+public class RecordServiceFactoryImpl implements RecordServiceFactory {
+
+  @Override
+  public RecordService create(Vertx vertx) {
+    return RecordServiceImpl.createForMultipleTables(
+      vertx,
+      Map.of(
+        "po_line",
+        List.of("po_line", "order_templates"),
+        "purchase_order",
+        List.of("purchase_order", "order_templates")));
+  }
+}
--- a/src/main/resources/META-INF/services/org.folio.service.spi.RecordServiceFactory
+++ b/src/main/resources/META-INF/services/org.folio.service.spi.RecordServiceFactory
@@ -1,0 +1,1 @@
+org.folio.services.record.RecordServiceFactoryImpl
--- a/src/main/resources/templates/db_scripts/order_templates_table.sql
+++ b/src/main/resources/templates/db_scripts/order_templates_table.sql
@@ -1,0 +1,2 @@
+CREATE INDEX IF NOT EXISTS order_templates_customfields_recordservice_idx_gin
+ON ${myuniversity}_${mymodule}.order_templates USING GIN ((jsonb->'customFields'));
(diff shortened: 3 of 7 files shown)


src/main/java/org/folio/services/record/RecordServiceFactoryImpl.java
```diff
@@ -0,0 +1,24 @@
...
+
+  @Override
+  public RecordService create(Vertx vertx) {
+    return RecordServiceImpl.createForMultipleTables(
+      vertx,
+      Map.of(
+        "po_line",
+        List.of("po_line", "order_templates"),
+        "purchase_order",
+        List.of("purchase_order", "order_templates")));
```
SerhiiNosko, line 22: It looks like this implementation makes many calls in orders tables with this query: WHERE jsonb->'customFields' ? $1
In this case can you check if index was added for custom fields that will cover this query?
You can check example how to add indexes in schema.json.
So yo ujust need to update that file and RMB will handle index creating

alb3rtino: Good point. I'll check.

alb3rtino: Creating indexes using RMB seems unsuitable in this case.
Adding to `schema.json` for table `purchase_order`:
```
"ginIndex": [
        {
          "fieldName": "customFields",
          "caseSensitive": true,
          "removeAccents": false
        }
      ]
```
I was only able to create the following index: 
```
"purchase_order_customfields_idx_gin" gin ((jsonb ->> 'customFields'::text) gin_trgm_ops)
```

Which does not cover the query.

Instead I manually added a `gin ((jsonb -> 'customFields'::text))` index to every table effected by this PR - `purchase_order`, `po_line` and `order_templates`.

explain_analyze

alb3rtino: The index creation worked with RMB, but the created index does not cover queries that use the `?` JSONB operator.

SerhiiNosko: can you also please ask this question in #raml-module-builder slack channel, if this is not supported, we will go with approach to manually create indexes as implemented now in the PR

alb3rtino: I can do that.

SerhiiNosko (approved):

damien-git (approved):

PR #390, opened 2024-03-19
[MODORDSTOR-364] Create routing_list table and API for it

Saba-Zedginidze-EPAM:
## Purpose
[MODORDSTOR-364: Create routing_list table and API for it](https://folio-org.atlassian.net/browse/MODORDSTOR-364)

## Approach
* Add a new table schema for routing lists
* Add to acq models: RoutingList, RoutingListCollection
* Add API for all CRUD operations

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -634,6 +634,37 @@
           "delay": "1"
         }
       ]
+    },
+    {
+      "id": "orders-storage.routing-lists",
+      "version": "1.0",
+      "handlers": [
+        {
+          "methods": ["GET"],
+          "pathPattern": "/orders-storage/routing-lists",
+          "permissionsRequired": ["orders-storage.routing-lists.collection.get"]
+        },
+        {
+          "methods": ["POST"],
+          "pathPattern": "/orders-storage/routing-lists",
+          "permissionsRequired": ["orders-storage.routing-lists.item.post"]
+        },
+        {
+          "methods": ["GET"],
+          "pathPattern": "/orders-storage/routing-lists/{id}",
+          "permissionsRequired": ["orders-storage.routing-lists.item.get"]
+        },
+        {
+          "methods": ["PUT"],
+          "pathPattern": "/orders-storage/routing-lists/{id}",
+          "permissionsRequired": ["orders-storage.routing-lists.item.put"]
+        },
+        {
+          "methods": ["DELETE"],
+          "pathPattern": "/orders-storage/routing-lists/{id}",
+          "permissionsRequired": ["orders-storage.routing-lists.item.delete"]
+        }
+      ]
     }
   ],
   "permissionSets" : [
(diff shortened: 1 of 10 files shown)


descriptors/ModuleDescriptor-template.json
```diff
@@ -634,6 +634,37 @@
           "delay": "1"
         }
       ]
+    },
```
SerhiiNosko, line 637: lets move this declaration after custom-fields before system interfaces _tenant, _timer

src/main/resources/templates/db_scripts/tables/create_routing_list_table.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE routing_list
```
SerhiiNosko, line 1: this is not needed, it's enough just to declare table in schema json.

ramls/routing-list.raml
```diff
@@ -0,0 +1,81 @@
+#%RAML 1.0
+title: "mod-orders-storage"
+baseUri: http://github.com/folio-org/mod-orders-storage
+version: v1.0
+
+documentation:
+  - title: Routing List
+    content: <b>This module implements the CRUD interface. This API is intended for internal use only. </b>
```
SerhiiNosko, line 8: We don't need to specify here "This API is intended for internal use only."
Also need to add info that this API is for routing lists

ramls/routing-list.raml
```diff
@@ -0,0 +1,81 @@
...
+      401:
+        description: "Not authorized to perform requested action"
+        body:
+          text/plain:
+            example: "Unable to add routing-list -- unauthorized"
+      500:
+        description: "Internal server error, e.g. due to misconfiguration"
+        body:
+          text/plain:
+            example: "Internal server error, contact administrator"
```
SerhiiNosko, line 69: for new APIs we should not provide text/plain response, we should provide application/json if some error is occurs, use the same declaration as for other APIs without text/plain

Abdulkhakimov (approved):

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -687,6 +687,25 @@
           "removeAccents": true
         }
       ]
+    },
+    {
+      "tableName": "routing_list",
+      "fromModuleVersion": "mod-orders-storage-13.8.0",
+      "withMetadata": true,
```
SerhiiNosko, line 694: there are chances that the same routing list can be edited by multiple users simultaniously, in the project we use optimistic locking technic to negotiate with this. We add optimistic locking for new tables and also have feature in backlog to add it to an existing tables.
So need to do 2 minor changes here: add "withOptimisticLocking": "failOnConflict" here and also add _version property to routing list schema.
You can use this PR as example: https://github.com/folio-org/mod-organizations-storage/pull/102/files

Saba-Zedginidze-EPAM: Added both

SerhiiNosko (approved):

PR #405, opened 2024-04-23
MODORDSTOR-399. Missing interface dependencies in module descriptor

SerhiiNosko:
## Purpose
https://folio-org.atlassian.net/browse/MODORDSTOR-399

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -661,10 +661,7 @@
       "handlers": [
         {
           "methods": ["POST"],
-          "pathPattern": "/_/tenant",
-          "modulePermissions": [
-            "finance-storage.funds.collection.get"
-          ]
+          "pathPattern": "/_/tenant"
         },
         {
           "methods": ["GET", "DELETE"],
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -698,6 +698,12 @@
       ]
     }
   ],
+  "requires": [
+    {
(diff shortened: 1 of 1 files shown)


descriptors/ModuleDescriptor-template.json
```diff
@@ -698,6 +698,12 @@
       ]
     }
   ],
+  "requires": [
+    {
+      "id": "configuration",
+      "version": "2.0"
+    }
```
azizbekxm, line 705: are we going to add finance storage too https://github.com/folio-org/mod-orders-storage/blob/80a80ecf09b8f90fe6c6d64e9193d099a6e81fd8/descriptors/ModuleDescriptor-template.json#L666

SerhiiNosko: Here is releted ticket for this: https://folio-org.atlassian.net/browse/MODORDSTOR-213
In TenantReferenceAPI I see that we don't have calls to finance-storage, also I don't see this SELECT statement that was described in that ticket:
SELECT jsonb -> 'code'
FROM ${myuniversity}_mod_finance_storage.fund

I think we can just delete that module permission

azizbekxm (approved):

descriptors/ModuleDescriptor-template.json
SerhiiNosko: Here we had ticket to remove finance-storage.funds dependency:
https://folio-org.atlassian.net/browse/MODORDSTOR-226
And module permission was not removed at that time, so removing now, thanks @azizbekxm for this catch

PR #412, opened 2024-04-26
[MODORDSTOR-364] Remove ID as required field from Routing Lists

Saba-Zedginidze-EPAM:
## Purpose
[[MODORDSTOR-364] Create routing_list table and API for it](https://folio-org.atlassian.net/browse/MODORDSTOR-364)

## Approach
- Remove ID field as required

### DIFF
--- a/src/main/java/org/folio/rest/impl/RoutingListsAPI.java
+++ b/src/main/java/org/folio/rest/impl/RoutingListsAPI.java
@@ -31,9 +31,6 @@ public class RoutingListsAPI implements OrdersStorageRoutingLists {
   @Validate
   public void postOrdersStorageRoutingLists(RoutingList entity, Map<String, String> okapiHeaders,
                                             Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
-    if (StringUtils.isBlank(entity.getId())) {
-      entity.setId(UUID.randomUUID().toString());
-    }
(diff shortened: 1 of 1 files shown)


SerhiiNosko (approved):

src/main/java/org/folio/rest/impl/RoutingListsAPI.java
```diff
@@ -31,9 +31,6 @@ public void getOrdersStorageRoutingLists(String query, String totalRecords, int
   @Validate
   public void postOrdersStorageRoutingLists(RoutingList entity, Map<String, String> okapiHeaders,
                                             Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
-    if (StringUtils.isBlank(entity.getId())) {
-      entity.setId(UUID.randomUUID().toString());
-    }
```
azizbekxm, line 36: after removing this part, since we deleted required field, where uuid is being generated.

azizbekxm: I mean in case of FE don't send generated uuid

Saba-Zedginidze-EPAM: PgClient does it

azizbekxm (approved):

PR #417, opened 2024-06-14
MODORDSTOR-392 Add index management for custom fields

alb3rtino:
# Purpose
https://folio-org.atlassian.net/browse/MODORDSTOR-392

This PR introduces dynamic index management for custom fields using database triggers. The dynamic nature of custom fields and the way that SQL queries are constructed prevents defining search indexes for custom fields in advance. This enhancement ensures the presence of search indexes for custom fields and thereby enhances the performance of search and filter operations related to custom fields.

# Approach
* Search indexes are automatically created when new custom fields are added.
* Search indexes are automatically dropped when custom fields are removed.
* Targets `purchase_order` and `po_line` tables.

# Screencast
[screencast.webm](https://github.com/folio-org/mod-orders-storage/assets/25431636/82471529-d56b-4ec8-a8fe-ecda1d9b66a5)

The screencast demonstrates the following functionalities for the `purchase_order` table: 
* Index creation upon adding custom fields.
* Utilization of indexes during search and filter operations.
* Index removal upon deletion of custom fields.

# Notes
Warnings about missing indexes for custom fields may still appear in the application log. This is because RMB determines the presence of an index based on the `schema.json` definition, which does not account for dynamically created indexes.


SerhiiNosko:
very smart implementation, thanks, approved

SerhiiNosko (approved):

PR #421, opened 2024-06-21
MODORDSTOR-408: Upgrading tenant resets reference and sample records

julianladisch:
https://folio-org.atlassian.net/browse/MODORDSTOR-408

Skip the installation of reference data and sample data if they have been installed in a previous version.

This way data that the institution has changed or deleted doesn't get reverted to the pristine values.

## Purpose
When migrating a tenant don't revert manual adjustments that implementers have made to the reference and sample data.

## Approach
When migrating from a mod-orders-storage version that already have installed reference or sample data then skip any task that might install or overwrite that data.

## Pre-Merge Checklist:


BKadirkhodjaev (approved):

SerhiiNosko (approved):

src/main/java/org/folio/rest/impl/TenantReferenceAPI.java
```diff
@@ -60,13 +61,19 @@ public Future<Integer> loadData(TenantAttributes attributes, String tenantId, Ma
       .onFailure(throwable -> Future.failedFuture(throwable.getCause()));
   }
 
+  private void buildDataLoadingParameters(TenantAttributes tenantAttributes, TenantLoading tl) {
+    if (isNew(tenantAttributes, "13.5.0")) {
```
SerhiiNosko, line 65: we use 13.5.0 here because reasons-for-closure OR acquisition-methods introduced in 13.5.0?

PR #432, opened 2024-09-20
[MODORDSTOR-417] - Add kafka consumer for Item Create events with processing logic

azizbekxm:
## Purpose
https://folio-org.atlassian.net/browse/MODORDSTOR-417

## Approach
- Refactored current Consumer Verticle for multiple consumers
- Implemented consumer for Item Create
- Implemented logic to fetch and update piece servie
- Implemented logic to update pieces using item fields, specifically tenantId and holdingRecordId


src/main/java/org/folio/event/handler/ItemCreateAsyncRecordHandler.java
```diff
@@ -0,0 +1,95 @@
...
+
+    var holdingId = itemObject.getString("holdingsRecordId");
+    updatePieceFields(pieces, holdingId, client.getTenantId());
+    return pieceService.updatePieces(pieces, client);
+  }
+
+  private void updatePieceFields(List<Piece> pieces, String holdingId, String tenantId) {
+    pieces.forEach(piece -> {
+      piece.setReceivingTenantId(tenantId);
+      piece.setHoldingId(holdingId);
```
SerhiiNosko, line 83: if tenantId and holdigId were not changed we should send a PUT request.
Also pieces can have only locationId populated. In this case we should skip updating holdigId. Only single field: lcoationId or holdingId can be populated in the same time

src/main/java/org/folio/verticles/KafkaConsumersVerticle.java
```diff
@@ -53,32 +56,111 @@ public KafkaConsumersVerticle(KafkaConfig kafkaConfig,
...
     context.put("springContext", springContext);
-    ediExportOrdersHistoryKafkaHandler = new EdiExportOrdersHistoryAsyncRecordHandler(context, vertx);
   }
 
   @Override
   public void start(Promise<Void> startPromise) {
-    log.info("Kafka config: {}", kafkaConfig);
+    log.info("start:: starting kafka consumer verticle. kafka config: {}", kafkaConfig);
+
+    createConsumers()
```
SerhiiNosko, line 65: not sure about usign single verticle for multiple caonsumers.

Each verticle in Vert.x typically runs on its own event loop thread. By deploying each Kafka consumer in a separate verticle, you take full advantage of the multicore architecture of your system, as each consumer can potentially run on a separate thread, achieving true parallel processing of consumed messages.

Can you search please in our github, I think we do not have cases when reusing the same verticle for multiple consumers

azizbekxm: Looked at other implementation and found that they also used similar approach, one verticle for all consumers. I think vertx run AbstractVerticle that is partent class for all consumer verticle
https://github.com/folio-org/mod-circulation-storage/blob/28aa0b72a3e651c8e5fccfa0ad8f7a222287d27e/src/main/java/org/folio/EventConsumerVerticle.java#L38
https://github.com/folio-org/mod-orders/blob/3eefce4d7ffd27f328bba23f974f1499ca81ee7b/src/main/java/org/folio/verticle/DataImportConsumerVerticle.java#L73
https://github.com/folio-org/mod-source-record-manager/blob/95da2781b71d0423604b866ac7ee0f91bbfe20ac/mod-source-record-manager-server/src/main/java/org/folio/verticle/AbstractConsumersVerticle.java#L48

src/main/java/org/folio/event/handler/ItemCreateAsyncRecordHandler.java
```diff
@@ -0,0 +1,113 @@
...
+    var updateRequiredPieces = filterPiecesToUpdate(pieces, holdingId, tenantId);
+    updatePieceFields(updateRequiredPieces, holdingId, tenantId);
+    return pieceService.updatePieces(updateRequiredPieces, client);
+  }
+
+  private List<Piece> filterPiecesToUpdate(List<Piece> pieces, String holdingId, String tenantId) {
+    return pieces.stream()
+      .filter(piece ->
+        !StringUtils.equals(piece.getReceivingTenantId(), tenantId)
+          && (Objects.isNull(piece.getLocationId()) || !StringUtils.equals(piece.getLocationId(), holdingId)))
```
SerhiiNosko, line 92: (Objects.isNull(piece.getLocationId()) || !StringUtils.equals(piece.getLocationId(), holdingId))
should be 
(Objects.isNull(piece.getLocationId()) && !StringUtils.equals(piece.getHoldingId(), holdingId))

src/main/java/org/folio/event/handler/ItemCreateAsyncRecordHandler.java
```diff
@@ -0,0 +1,113 @@
...
+    var itemObject = payload.getJsonObject("new");
+    var itemId = itemObject.getString("id");
+
+    return pieceService.getPiecesByItemId(itemId, dbClient)
+      .compose(pieces -> updatePieces(pieces, itemObject, dbClient));
+  }
+
+  private Future<Void> updatePieces(List<Piece> pieces, JsonObject itemObject, DBClient client) {
+    if (CollectionUtils.isEmpty(pieces)) {
+      log.info("updatePieces:: no pieces to update found, nothing to update");
```
SerhiiNosko, line 77: this log is not very useful now, will be useful after adding identifier, for example itemId

src/main/java/org/folio/event/handler/ItemCreateAsyncRecordHandler.java
SerhiiNosko: by the way, this check with holdingId will not help in scope of this story, because holding ids would be the same after changing ownership to another tenant, because we have this story: https://folio-org.atlassian.net/browse/MODINV-1074 

But this check with holdingId will be useful for our story from next sprint when we chaning holding acrross single tenant. In this case holding ids would be different

src/main/java/org/folio/event/handler/ItemCreateAsyncRecordHandler.java
```diff
@@ -0,0 +1,113 @@
...
+      return Future.succeededFuture();
+    }
+
+    var holdingId = itemObject.getString("holdingsRecordId");
+    var tenantId = client.getTenantId();
```
SerhiiNosko, line 82: we need to get tenantId from kafka message headers or payload(if available in payload).

src/main/java/org/folio/verticles/KafkaConsumersVerticle.java
SerhiiNosko: okay got it, thinking what is better have 2 verticles:  Item Create, Item update and (Item delete in future), and Holding Create, Holding Update and (Holding Delete in future) events or use single verticle.
More like approach with 2 verticles(one all holdings events, another for all item events) to have better parallelism I think.
For mod-users we even have 3 separate verticles for the same entity: https://github.com/folio-org/mod-users/tree/master/src/main/java/org/folio/verticle

src/main/java/org/folio/event/handler/ItemCreateAsyncRecordHandler.java
azizbekxm: Thanks, update filter to skip same tenantId, holdingId or existingLocationId

src/main/java/org/folio/event/handler/ItemCreateAsyncRecordHandler.java
azizbekxm: Thanks added extra details

src/main/java/org/folio/event/handler/ItemCreateAsyncRecordHandler.java
azizbekxm: Actually in hanlde method tenantId is being set to dbClient from headers `dbClient.setTenant(tenantId)`. However this is seems easy to miss, changed implementation to get from kafka message.

src/main/java/org/folio/event/handler/ItemCreateAsyncRecordHandler.java
```diff
@@ -0,0 +1,116 @@
...
+        .onSuccess(v -> log.info("ItemCreateAsyncRecordHandler::handle, event processed successfully"))
+        .onFailure(t -> log.error("Failed to process event: {}", kafkaConsumerRecord.value(), t))
+        .map(kafkaConsumerRecord.key());
+    } catch (Exception e) {
+      log.error("Failed to process export history kafka record, kafkaRecord={}", kafkaConsumerRecord, e);
+      return Future.failedFuture(e);
+    }
+  }
+
+  private Future<Void> processItemCreationEvent(JsonObject payload, DBClient dbClient) {
```
BKadirkhodjaev, line 67: Could've been a protected or package-private method & tested without any reflections

BKadirkhodjaev (approved):

src/main/java/org/folio/event/handler/ItemCreateAsyncRecordHandler.java
```diff
@@ -0,0 +1,116 @@
...
+    var payload = new JsonObject(kafkaConsumerRecord.value());
+
+    String eventType = payload.getString("type");
+    if (!Objects.equals(eventType, DomainEventPayloadType.CREATE.name())) {
+      log.info("processItemCreateEvent:: unsupported event type: {}", eventType);
+      return Future.succeededFuture();
+    }
+
+    if (!validatePayload(payload)) {
+      log.warn("processItemCreateEvent:: payload validation failed");
```
SerhiiNosko, line 50: payload can be added to warn log

src/main/java/org/folio/event/handler/ItemCreateAsyncRecordHandler.java
```diff
@@ -0,0 +1,117 @@
...
+
+    try {
+      var tenantId = extractTenantFromHeaders(kafkaConsumerRecord.headers());
+      var dbClient = new DBClient(getVertx(), tenantId);
+      return processItemCreationEvent(payload, dbClient)
+        .onSuccess(v -> log.info("ItemCreateAsyncRecordHandler::handle, event processed successfully"))
+        .onFailure(t -> log.error("Failed to process event: {}", kafkaConsumerRecord.value(), t))
+        .map(kafkaConsumerRecord.key());
+    } catch (Exception e) {
+      log.error("Failed to process export history kafka record, kafkaRecord={}", kafkaConsumerRecord, e);
```
SerhiiNosko, line 62: log message need to be updated

src/main/java/org/folio/event/handler/ItemCreateAsyncRecordHandler.java
```diff
@@ -0,0 +1,117 @@
...
+    return pieceService.updatePieces(updateRequiredPieces, client);
+  }
+
+  private List<Piece> filterPiecesToUpdate(List<Piece> pieces, String holdingId, String tenantId) {
+    return pieces.stream()
+      .filter(piece ->
+        // filter out pieces that already have the same tenantId and (same holdingId or existing locationId)
+        ObjectUtils.notEqual(piece.getReceivingTenantId(), tenantId)
+          && Objects.isNull(piece.getLocationId())
+          && ObjectUtils.notEqual(piece.getHoldingId(), holdingId))
```
SerhiiNosko, line 96: hodlingId will be the same if we just switch it from tenantA to tenantB. So need to delete this check

SerhiiNosko: In scope of this story https://folio-org.atlassian.net/browse/MODORDSTOR-419 we will need it , because item can be moved to another holding in scope of the same tenant.
But for that check we will use Item UPDATE event, not create one

src/main/java/org/folio/event/handler/ItemCreateAsyncRecordHandler.java
```diff
@@ -0,0 +1,117 @@
...
+      .toList();
+  }
+
+  private void updatePieceFields(List<Piece> pieces, String holdingId, String tenantId) {
+    pieces.forEach(piece -> {
+      piece.setReceivingTenantId(tenantId);
+      if (Objects.isNull(piece.getLocationId())) {
```
SerhiiNosko, line 103: we can skip this if statement t oset holdingId, because for Item Create event holdingId could not be changed. It can be changed for Item Update event

src/main/java/org/folio/event/handler/ItemCreateAsyncRecordHandler.java
```diff
@@ -0,0 +1,117 @@
...
+  @Override
+  public Future<String> handle(KafkaConsumerRecord<String, String> kafkaConsumerRecord) {
+    if (log.isDebugEnabled()) {
+      log.debug("ItemCreateAsyncRecordHandler::handle, kafkaRecord={}", kafkaConsumerRecord.value());
+    }
+
+    var payload = new JsonObject(kafkaConsumerRecord.value());
+
+    String eventType = payload.getString("type");
+    if (!Objects.equals(eventType, DomainEventPayloadType.CREATE.name())) {
```
SerhiiNosko, line 44: thinking that we can delete these 2 checks, because if we subscribe to Create kafka topic we always will fetch create events.
And for validation the same, its core part to have the same contract for inventory-storage producer, but as you prefer, you can keep

src/main/java/org/folio/event/handler/ItemCreateAsyncRecordHandler.java
azizbekxm: updated, thanks to notice

src/main/java/org/folio/event/handler/ItemCreateAsyncRecordHandler.java
azizbekxm: good idea, added payload

src/main/java/org/folio/event/handler/ItemCreateAsyncRecordHandler.java
azizbekxm: I prefer extra checks without dependent on other modules. The reason is that the logic can be changed, in this case this part is working without know main logic of process and can fix it. in my opinion It is useful

src/main/java/org/folio/event/handler/ItemCreateAsyncRecordHandler.java
azizbekxm: Updated test with different cases, in any case new filter avoid unepxected cases even if it is not possible, still from my perspective, it is good to have some filter can work without dependent on other logic of other modules

src/main/java/org/folio/event/handler/ItemCreateAsyncRecordHandler.java
azizbekxm: I consider this method should be invoked on from one places and under specific cases, so private good for me and I usually avoid to change method beacuse of test:). for know I don't know any significant benefit from not using reflections. However, I am open to change if this strongly need

src/main/java/org/folio/verticles/KafkaConsumersVerticle.java
azizbekxm: Seems this style is a little bit confusing. Behind the scenes, this structure is using same verticle for all like current approach. Extra Verticles providing specific topic and handler. From my opinion verticles in mod-users are overengineered by having three different verticle for just prodiving different topic and handles. I have doubt it can achieve any performance benefit from this

Regarding your opinion, it make sense to have different verticles for different kafka topic, one for inventory one for export for better clarify.

src/main/java/org/folio/event/handler/ItemCreateAsyncRecordHandler.java
SerhiiNosko: ok got it, thanks

src/main/java/org/folio/event/handler/ItemCreateAsyncRecordHandler.java
```diff
@@ -0,0 +1,121 @@
...
+import org.folio.event.DomainEventPayloadType;
+import org.folio.rest.jaxrs.model.Piece;
+import org.folio.rest.persist.DBClient;
+import org.folio.services.piece.PieceService;
+import org.folio.spring.SpringContextUtil;
+import org.springframework.beans.factory.annotation.Autowired;
+
+public class ItemCreateAsyncRecordHandler extends BaseAsyncRecordHandler<String, String> {
+
+  private static final Logger log = LogManager.getLogger();
```
Saba-Zedginidze-EPAM, line 25: We can use `@Log4j` / `@Log4j2` annotation from lombok to generate loggers
https://projectlombok.org/features/log

src/main/java/org/folio/event/handler/ItemCreateAsyncRecordHandler.java
```diff
@@ -0,0 +1,121 @@
...
+  private PieceService pieceService;
+
+  public ItemCreateAsyncRecordHandler(Context context, Vertx vertx) {
+    super(vertx, context);
+    SpringContextUtil.autowireDependencies(this, context);
+  }
+
+  @Override
+  public Future<String> handle(KafkaConsumerRecord<String, String> kafkaConsumerRecord) {
+    if (log.isDebugEnabled()) {
```
Saba-Zedginidze-EPAM, line 37: Probably don't need this check as message evaluation is not expensive, only calling `kafkaConsumerRecord.value()`

src/main/java/org/folio/event/handler/ItemCreateAsyncRecordHandler.java
```diff
@@ -0,0 +1,121 @@
...
+    } catch (Exception e) {
+      log.error("Failed to process item event kafka record, kafkaRecord={}", kafkaConsumerRecord, e);
+      return Future.failedFuture(e);
+    }
+  }
+
+  private Future<Void> processItemCreationEvent(JsonObject payload, DBClient dbClient) {
+    var tenantId = payload.getString("tenant");
+    var itemObject = payload.getJsonObject("new");
+    var itemId = itemObject.getString("id");
```
Saba-Zedginidze-EPAM, line 70: Can be helpful to move the parameter values to static variables or even better option in an enum. Will be good for reusability

src/main/java/org/folio/event/handler/ItemCreateAsyncRecordHandler.java
```diff
@@ -0,0 +1,121 @@
...
+
+  private List<Piece> filterPiecesToUpdate(List<Piece> pieces, String holdingId, String tenantId) {
+    return pieces.stream()
+      .filter(piece ->
+        ObjectUtils.notEqual(piece.getReceivingTenantId(), tenantId)
+        || ObjectUtils.notEqual(piece.getHoldingId(), holdingId))
+      .filter(piece ->
+        ObjectUtils.notEqual(piece.getReceivingTenantId(), tenantId)
+        || Objects.isNull(piece.getLocationId()))
+      .toList();
```
Saba-Zedginidze-EPAM, line 101: Suggest to rewrite as similar to this with less repeating
```java
    return pieces.stream()
      .filter(piece -> ObjectUtils.notEqual(piece.getReceivingTenantId(), tenantId)
        || ObjectUtils.notEqual(piece.getHoldingId(), holdingId))
        || Objects.isNull(piece.getLocationId()))
      .toList();
```

src/main/java/org/folio/event/handler/ItemCreateAsyncRecordHandler.java
```diff
@@ -0,0 +1,121 @@
...
+    pieces.forEach(piece -> {
+      piece.setReceivingTenantId(tenantId);
+      if (Objects.isNull(piece.getLocationId())) {
+        piece.setHoldingId(holdingId);
+      }
+    });
+  }
+
+  private boolean validatePayload(JsonObject payload) {
+    var newObject = payload.getJsonObject("new");
```
Saba-Zedginidze-EPAM, line 114: We can avoid creating additional variable here

src/main/java/org/folio/event/EdiExportHistoryEventType.java
```diff
@@ -5,8 +5,9 @@
 
 @RequiredArgsConstructor
 @Getter
-public enum ExportEventType {
-  EXPORT_HISTORY_CREATE("edi-export-history.create");
+public enum EdiExportHistoryEventType {
+  EXPORT_HISTORY_CREATE("edi-export-history.create"),
+  INVENTORY_ITEM_CREATE("inventory.item");
```
SerhiiNosko, line 10: INVENTORY_ITEM_CREATE should not be in EdiExportHistoryEventType

SerhiiNosko (approved):

src/main/java/org/folio/event/EdiExportHistoryEventType.java
azizbekxm: Thanks, removed

src/main/java/org/folio/event/handler/ItemCreateAsyncRecordHandler.java
azizbekxm: Remove validatoin and moved to handle method since it is small, thanks to notice

src/main/java/org/folio/event/handler/ItemCreateAsyncRecordHandler.java
azizbekxm: Tested different approach but filter don't work and filter out valid cases, so keeping exisitng one Added comment

src/main/java/org/folio/event/handler/ItemCreateAsyncRecordHandler.java
azizbekxm: Agreee, created enum and dto for these

src/main/java/org/folio/event/handler/ItemCreateAsyncRecordHandler.java
azizbekxm: removed, thanks

src/main/java/org/folio/event/handler/ItemCreateAsyncRecordHandler.java
azizbekxm: Added to both hadnlers

PR #433, opened 2024-09-26
[MODORDSTOR-416] Add kafka consumer for Holdings Create events with processing logic

Saba-Zedginidze-EPAM:
## Purpose
[[MODORDSTOR-416] Add kafka consumer for Holdings Create events with processing logic
](https://folio-org.atlassian.net/browse/MODORDSTOR-416)

## Approach
- Add inventory consumer verticle and async handler to extend both with item and holding ones
- Add holding event async handler with processing logic
- Add holding consumer verticle 
- Add ConsortiumConfigurationService (same as in mod-orders)
- Fetch pieces and po lines from central tenant in case of ECS
- Update unit tests

### DIFF
--- a/src/main/java/org/folio/dao/lines/PoLinesPostgresDAO.java
+++ b/src/main/java/org/folio/dao/lines/PoLinesPostgresDAO.java
@@ -64,27 +64,11 @@ public class PoLinesPostgresDAO implements PoLinesDAO {
 
   @Override
   public Future<Integer> updatePoLines(String sql, Conn conn) {
-    log.debug("updatePoLines, sql={}", sql);
-    Promise<Integer> promise = Promise.promise();
-    conn.execute(sql)
-      .onSuccess(result -> {
-        log.debug("updatePoLines success, sql={}", sql);
-        promise.complete(result.rowCount());
-      })
-      .onFailure(t -> {
-        log.error("updatePoLines failed, sql={}", sql, t);
-        handleFailure(promise, t);
-      });
-    return promise.future();
-  }
-
-  @Override
-  public Future<Integer> updatePoLines(String sql, DBClient dbClient) {
-    log.debug("updatePoLines, sql={}", sql);
-    return dbClient.getPgClient().execute(sql)
+    log.debug("updatePoLines:: sql={}", sql);
+    return conn.execute(sql)
       .map(SqlResult::rowCount)
-      .onSuccess(result -> log.debug("updatePoLines success, sql={}", sql))
-      .onFailure(t -> log.error("updatePoLines failed, sql={}", sql, t));
-
+      .onSuccess(result -> log.debug("updatePoLines:: success, sql={}", sql))
+      .onFailure(t -> log.error("updatePoLines:: failed, sql={}", sql, t));
   }
+
 }
--- a/src/main/java/org/folio/dao/lines/PoLinesPostgresDAO.java
+++ b/src/main/java/org/folio/dao/lines/PoLinesPostgresDAO.java
@@ -83,12 +83,8 @@ public class PoLinesPostgresDAO implements PoLinesDAO {
     log.debug("updatePoLines, sql={}", sql);
     return dbClient.getPgClient().execute(sql)
       .map(SqlResult::rowCount)
-      .onSuccess(result -> {
-        log.debug("updatePoLines success, sql={}", sql);
-      })
-      .onFailure(t -> {
-        log.error("updatePoLines failed, sql={}", sql, t);
-      });
+      .onSuccess(result -> log.debug("updatePoLines success, sql={}", sql))
+      .onFailure(t -> log.error("updatePoLines failed, sql={}", sql, t));
 
   }
 }
--- a/src/main/java/org/folio/dao/lines/PoLinesPostgresDAO.java
+++ b/src/main/java/org/folio/dao/lines/PoLinesPostgresDAO.java
@@ -18,6 +18,7 @@ import org.folio.rest.persist.Criteria.Criterion;
 import io.vertx.core.Future;
 import io.vertx.core.Promise;
 import io.vertx.core.json.JsonObject;
+import io.vertx.sqlclient.SqlResult;
 
 
 public class PoLinesPostgresDAO implements PoLinesDAO {
@@ -75,5 +76,19 @@ public class PoLinesPostgresDAO implements PoLinesDAO {
         handleFailure(promise, t);
       });
     return promise.future();
+  }
+
+  @Override
+  public Future<Integer> updatePoLines(String sql, DBClient dbClient) {
+    log.debug("updatePoLines, sql={}", sql);
+    return dbClient.getPgClient().execute(sql)
+      .map(SqlResult::rowCount)
+      .onSuccess(result -> {
+        log.debug("updatePoLines success, sql={}", sql);
+      })
+      .onFailure(t -> {
+        log.error("updatePoLines failed, sql={}", sql, t);
+      });
+
   }
 }
--- a/src/main/java/org/folio/event/handler/HoldingCreateAsyncRecordHandler.java
+++ b/src/main/java/org/folio/event/handler/HoldingCreateAsyncRecordHandler.java
@@ -8,14 +8,17 @@ import java.util.Objects;
 import java.util.Optional;
 
 import org.apache.commons.collections4.CollectionUtils;
-import org.apache.logging.log4j.Logger;
 import org.folio.event.dto.InventoryFields;
 import org.folio.event.dto.ResourceEvent;
+import org.folio.event.service.AuditOutboxService;
 import org.folio.models.ConsortiumConfiguration;
 import org.folio.okapi.common.GenericCompositeFuture;
 import org.folio.rest.jaxrs.model.Location;
+import org.folio.rest.jaxrs.model.OrderLineAuditEvent;
 import org.folio.rest.jaxrs.model.Piece;
+import org.folio.rest.jaxrs.model.PieceAuditEvent;
 import org.folio.rest.jaxrs.model.PoLine;
+import org.folio.rest.persist.Conn;
 import org.folio.rest.persist.DBClient;
 import org.folio.services.consortium.ConsortiumConfigurationService;
 import org.folio.services.lines.PoLinesService;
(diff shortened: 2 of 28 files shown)


src/main/java/org/folio/dao/lines/PoLinesPostgresDAO.java
```diff
@@ -76,4 +77,14 @@ public Future<Integer> updatePoLines(String sql, Conn conn) {
       });
     return promise.future();
   }
+
+  @Override
+  public Future<Integer> updatePoLines(String sql, DBClient dbClient) {
```
SerhiiNosko, line 82: we should not use DBClient in new code, better to use Conn entity, so you can refactor with usage of dbClient.withConn()
There are 2 reasons:

- previosly we had some issues espesially with using transactions when DBClient returns 2 different pg clients that returns different sql connections from the pool, so transaction rollback in this case will not work
- we need to write audit log for each po line update, so we need to write this audit log in scope of the same transaction that accepts Conn object

src/main/java/org/folio/event/handler/HoldingCreateAsyncRecordHandler.java
```diff
@@ -0,0 +1,99 @@
...
+    var holdingId = holdingObject.getString(InventoryFields.ID.getValue());
+    var dbClient = new DBClient(getVertx(), tenantId);
+    var tenantIdUpdates = List.of(
+      processPoLinesUpdate(holdingId, tenantId, dbClient),
+      processPiecesUpdate(holdingId, tenantId, dbClient)
+    );
+    return GenericCompositeFuture.all(tenantIdUpdates).mapEmpty();
+  }
+
+  private Future<Void> processPoLinesUpdate(String holdingId, String tenantId, DBClient dbClient) {
```
SerhiiNosko, line 54: we need to store info about each po line update in mod-audit, so here we should use producer to push messages to kafka. You can see how to use it by searching auditOutboxService.saveOrderLinesOutboxLogs

SerhiiNosko: So propose to open 1 transaction and use that connection for po line updates, piece updates and saving po line details into audit log

src/main/java/org/folio/event/handler/HoldingCreateAsyncRecordHandler.java
```diff
@@ -0,0 +1,99 @@
...
+    return poLinesService.getPoLinesByHoldingId(holdingId, dbClient)
```
SerhiiNosko, line 62: do not see index for location.holdingId in schema.json, better to add

src/main/java/org/folio/event/handler/HoldingCreateAsyncRecordHandler.java
```diff
@@ -0,0 +1,99 @@
...
+
+  private void updateLocationTenantIdIfNeeded(List<Location> locations, String holdingId, String tenantId) {
+    locations.stream()
+      .filter(location -> Objects.equals(location.getHoldingId(), holdingId))
+      .forEach(location -> location.setTenantId(tenantId));
+  }
+
+  @Override
+  protected Logger getLogger() {
+    return log;
```
SerhiiNosko, line 108: can we avoid this method?

src/main/java/org/folio/event/handler/HoldingCreateAsyncRecordHandler.java
Saba-Zedginidze-EPAM: Should we not do the same for Pieces also?

src/main/java/org/folio/event/handler/HoldingCreateAsyncRecordHandler.java
Saba-Zedginidze-EPAM: We can create logger in the parent class but it would appear as InventoryCreateAsyncRecordHandler in the logs instead of a concrete child class

SerhiiNosko: we can just declare separate loggers in each class

src/main/java/org/folio/event/handler/HoldingCreateAsyncRecordHandler.java
SerhiiNosko: It is good question, basically in mod-audit we have an endpoint only for finding pieces with changed statuses. But in mod-orders-storage we always send events for app pieces operations. I think we should be consistent and do the same here for pieces as well.

src/main/java/org/folio/event/handler/HoldingCreateAsyncRecordHandler.java
```diff
@@ -0,0 +1,118 @@
...
+    var holdingObject = JsonObject.mapFrom(resourceEvent.getNewValue());
+    var holdingId = holdingObject.getString(InventoryFields.ID.getValue());
+    return new DBClient(getVertx(), tenantId).getPgClient()
+      .withTrans(conn -> {
+        var tenantIdUpdates = List.of(
+          processPoLinesUpdate(holdingId, tenantId, headers, conn),
+          processPiecesUpdate(holdingId, tenantId, headers, conn)
+        );
+        return GenericCompositeFuture.all(tenantIdUpdates).mapEmpty();
+      });
```
SerhiiNosko, line 66: we need to add .onComplete section to process transaction outbox tables, see usages of "auditOutboxService.processOutboxEventLogs(okapiHeaders);"

src/main/java/org/folio/event/handler/HoldingCreateAsyncRecordHandler.java
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  private Future<Void> processPoLinesUpdate(String holdingId, String tenantId, Map<String, String> headers, Conn conn) {
+    return poLinesService.getPoLinesByHoldingId(holdingId, conn)
+      .compose(poLines -> updatePoLines(poLines, holdingId, tenantId, conn))
+      .compose(poLines -> auditOutboxService.saveOrderLinesOutboxLogs(conn, poLines, OrderLineAuditEvent.Action.EDIT, headers))
+      .onSuccess(ar -> auditOutboxService.processOutboxEventLogs(headers))
```
SerhiiNosko, line 73: we can invoke this method to process outbox event logs only once after transaction commited. By the way this pattern is used here: https://microservices.io/patterns/data/transactional-outbox.html

src/main/java/org/folio/event/handler/HoldingCreateAsyncRecordHandler.java
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  private void updateLocationTenantIdIfNeeded(List<Location> locations, String holdingId, String tenantId) {
+    locations.stream()
+      .filter(location -> Objects.equals(location.getHoldingId(), holdingId))
+      .forEach(location -> location.setTenantId(tenantId));
+  }
+
+  @Override
+  protected Future<Optional<ConsortiumConfiguration>> getConsortiumConfiguration(Map<String, String> headers) {
```
SerhiiNosko, line 116: this method can be just in base class, because it has the same implementation in subclasses

src/main/java/org/folio/event/handler/ItemCreateAsyncRecordHandler.java
```diff
@@ -1,90 +1,74 @@
...
     var itemObject = JsonObject.mapFrom(resourceEvent.getNewValue());
-    var itemId = itemObject.getString(ItemField.ID.getValue());
-
-    return pieceService.getPiecesByItemId(itemId, dbClient)
-      .compose(pieces -> updatePieces(pieces, itemObject, tenantId, dbClient));
+    var itemId = itemObject.getString(InventoryFields.ID.getValue());
+    return new DBClient(getVertx(), tenantId).getPgClient()
+      .withTrans(conn -> pieceService.getPiecesByItemId(itemId, conn)
+        .compose(pieces -> updatePieces(pieces, itemObject, tenantId, conn))
+        .compose(pieces -> auditOutboxService.savePiecesOutboxLog(conn, pieces, PieceAuditEvent.Action.CREATE, headers))
```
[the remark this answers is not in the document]
Saba-Zedginidze-EPAM: Yes that was the initial thought, thanks for noticing, will fix

src/main/java/org/folio/event/handler/InventoryCreateAsyncRecordHandler.java
```diff
@@ -0,0 +1,73 @@
...
+        .onSuccess(v -> log.info("handle:: '{}' event for '{}' processed successfully", eventType, inventoryEventType.getTopicName()))
+        .onFailure(t -> log.error("Failed to process event: {}", recordValue, t))
+        .map(kafkaConsumerRecord.key());
+    } catch (Exception e) {
+      log.error("Failed to process kafkaConsumerRecord: {}", kafkaConsumerRecord, e);
+      return Future.failedFuture(e);
+    }
+  }
+
+  private Future<String> getTenantId(Map<String, String> headers) {
```
SerhiiNosko, line 62: should be used for both Item Create and Holding Create. So we also need to update po lines/pieces from central tenant for ECS envs.
Also some log can be added that tenantId is ovveriden, because it is not very obvious.

src/main/java/org/folio/event/handler/HoldingCreateAsyncRecordHandler.java
Saba-Zedginidze-EPAM: Thought we needed to process the outbox logs separately as one could fail but now I see we should revert everything altogether so will move this after the completion of whole tx

src/main/java/org/folio/event/handler/InventoryCreateAsyncRecordHandler.java
Saba-Zedginidze-EPAM: It is used for both, InventoryCreateAsyncRecordHandler is the parent class for both item and holding handlers. Will add some logging

src/main/java/org/folio/event/handler/InventoryCreateAsyncRecordHandler.java
```diff
@@ -0,0 +1,87 @@
...
```
SerhiiNosko, line 67: one more optimization that we can apply here is to skip making requests to DB to fetch po lines, pieces in case when consortium is not set up. For items it can be millions of them and in this case we will reduce select calls to DB.

Saba-Zedginidze-EPAM: Correct, we only need to process in ecs envs, thanks.

src/main/java/org/folio/event/handler/InventoryCreateAsyncRecordHandler.java
```diff
@@ -69,10 +71,10 @@ private Future<String> getTenantId(Map<String, String> headers) {
     return consortiumConfigurationService.getConsortiumConfiguration(headers)
       .map(optionalConsortiumConfiguration -> optionalConsortiumConfiguration
         .map(ConsortiumConfiguration::centralTenantId)
-        .orElse(headersTenantId))
+        .orElseThrow(() -> new IllegalStateException(CONSORTIUM_CONFIG_NOT_FOUND)))
```
SerhiiNosko, line 74: we need to rewrite logic to avoid building it on throwing/catching exceptions.
See more at chapter: "Use exceptions only for exceptional conditions." of Effective Java
And also having errors in logs can confuse that something went wrong. You can just return from this method centralTenantId and from the upper layer in base class check that if incoming kafka message is not from central tenant, just skip processing, probably even debug level log would be enough.

Saba-Zedginidze-EPAM: Thanks for suggestion, sounds better to get rid of exception. However I think we should not consider headers' tenantId as to be a failed scenario, as it could be the same as central tenant id. Better to set it null and then to the null-check before processing

SerhiiNosko: looks good now

SerhiiNosko (approved):
Looks very good

src/main/java/org/folio/verticles/InventoryConsumersVerticle.java
```diff
@@ -0,0 +1,52 @@
...
+   * This method creates a consumer for the given event type.
+   * Note: The method is using specific subscription pattern for inventory topics:
+   * {envId}.{tenant}.{eventType} -> e.g. 'folio.diku.inventory.item'
+   *
+   * @return future with the created consumer
+   */
+  @Override
+  protected Future<Void> createConsumers() {
+    var subscriptionPattern = KafkaTopicNameHelper.formatTopicName(kafkaConfig.getEnvId(), TENANT_ID_PATTERN, eventType.getTopicName());
+    log.info("createConsumers:: Creating Inventory Event consumers with subscriptionPattern: {} for evenType: {}", subscriptionPattern, eventType.name());
```
azizbekxm, line 43: these look like a long line, it makes hard to read

Saba-Zedginidze-EPAM: Tried shortening some lines, should be good now

src/main/java/org/folio/event/handler/ItemCreateAsyncRecordHandler.java
```diff
@@ -1,90 +1,67 @@
...
     var itemObject = JsonObject.mapFrom(resourceEvent.getNewValue());
-    var itemId = itemObject.getString(ItemField.ID.getValue());
-
-    return pieceService.getPiecesByItemId(itemId, dbClient)
-      .compose(pieces -> updatePieces(pieces, itemObject, tenantId, dbClient));
+    var itemId = itemObject.getString(InventoryFields.ID.getValue());
+    return dbClient.getPgClient()
+      .withTrans(conn -> pieceService.getPiecesByItemId(itemId, conn)
+        .compose(pieces -> updatePieces(pieces, itemObject, tenantId, conn))
+        .compose(pieces -> auditOutboxService.savePiecesOutboxLog(conn, pieces, PieceAuditEvent.Action.CREATE, headers)))
```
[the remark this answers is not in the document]
Saba-Zedginidze-EPAM: Thanks for noticing, fixed in both handlers

src/main/java/org/folio/event/handler/InventoryCreateAsyncRecordHandler.java
```diff
@@ -0,0 +1,92 @@
...
+        .onSuccess(v -> log.info("handle:: '{}' event for '{}' processed successfully", eventType, inventoryEventType.getTopicName()))
+        .onFailure(t -> log.error("Failed to process event: {}", recordValue, t))
+        .map(kafkaConsumerRecord.key());
+    } catch (Exception e) {
+      log.error("Failed to process kafkaConsumerRecord: {}", kafkaConsumerRecord, e);
+      return Future.failedFuture(e);
+    }
+  }
+
+  private Future<String> getTenantId(Map<String, String> headers) {
```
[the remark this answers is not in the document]
Saba-Zedginidze-EPAM: Good catch, renamed

PR #436, opened 2024-10-03
[MODORDSTOR-416] Skip kafka event processing without central ordering

Saba-Zedginidze-EPAM:
## Purpose
[[MODORDSTOR-416] Add kafka consumer for Holdings Create events with processing logic
](https://folio-org.atlassian.net/browse/MODORDSTOR-416)

## Approach
- Add SettingsService and use in controller and event handler
- Add add ResourcePath enum for declaring paths for RestClient

### DIFF
--- a/src/main/java/org/folio/event/handler/InventoryCreateAsyncRecordHandler.java
+++ b/src/main/java/org/folio/event/handler/InventoryCreateAsyncRecordHandler.java
@@ -10,10 +10,12 @@ import java.util.Objects;
 import org.apache.commons.collections4.map.CaseInsensitiveMap;
 import org.folio.event.InventoryEventType;
 import org.folio.event.dto.ResourceEvent;
-import org.folio.models.ConsortiumConfiguration;
+import org.folio.rest.jaxrs.model.Setting;
 import org.folio.rest.core.models.RequestContext;
 import org.folio.rest.persist.DBClient;
 import org.folio.services.consortium.ConsortiumConfigurationService;
+import org.folio.services.settings.SettingsService;
+import org.folio.services.settings.util.SettingKey;
 import org.springframework.beans.factory.annotation.Autowired;
 
 import io.vertx.core.Context;
@@ -28,6 +30,9 @@ public abstract class InventoryCreateAsyncRecordHandler extends BaseAsyncRecordH
 
   @Autowired
   protected ConsortiumConfigurationService consortiumConfigurationService;
+
+  @Autowired
+  protected SettingsService settingsService;
 
   private final InventoryEventType inventoryEventType;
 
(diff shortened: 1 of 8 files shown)


src/main/java/org/folio/services/setting/SettingService.java
```diff
@@ -0,0 +1,71 @@
...
+
+  public Future<Response> getSettings(String query, int offset, int limit, Map<String, String> okapiHeaders, Context vertxContext) {
+    return PgUtil.get(SETTINGS_TABLE, Setting.class, SettingCollection.class, query, offset, limit, okapiHeaders, vertxContext,
+      OrdersStorageSettings.GetOrdersStorageSettingsResponse.class);
+  }
+
+  public Future<Optional<Setting>> getSettingByKey(SettingKey settingKey, Map<String, String> okapiHeaders, Context vertxContext) {
+    return getSettings(String.format(SETTINGS_BY_KEY_QUERY, settingKey.getName()), 0, 1, okapiHeaders, vertxContext)
+      .map(response -> JsonObject.mapFrom(response.getEntity()).mapTo(SettingCollection.class))
+      .map(settings -> settings.getTotalRecords() == null || settings.getTotalRecords() != 1 || CollectionUtils.isEmpty(settings.getSettings())
```
BKadirkhodjaev, line 41: This is bit hard to read

src/main/java/org/folio/services/setting/SettingService.java
```diff
@@ -0,0 +1,71 @@
...
+      OrdersStorageSettings.PutOrdersStorageSettingsByIdResponse.class, asyncResultHandler);
+  }
+
+  public void deleteSetting(String id, Map<String, String> okapiHeaders,
+                            Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
+    PgUtil.deleteById(SETTINGS_TABLE, id, okapiHeaders, vertxContext,
+      OrdersStorageSettings.PutOrdersStorageSettingsByIdResponse.class, asyncResultHandler);
+  }
+
+
```
BKadirkhodjaev, line 70: Extra empty line

BKadirkhodjaev (approved):

src/main/java/org/folio/event/handler/InventoryCreateAsyncRecordHandler.java
```diff
@@ -86,9 +91,17 @@ private void verifyKafkaRecord(KafkaConsumerRecord<String, String> kafkaConsumer
...
     return consortiumConfigurationService.getConsortiumConfiguration(requestContext)
-      .map(optionalConsortiumConfiguration -> optionalConsortiumConfiguration
-        .map(ConsortiumConfiguration::centralTenantId)
-        .orElse(null));
+      .compose(consortiumConfiguration -> {
+        if (consortiumConfiguration.isEmpty()) {
+          return Future.succeededFuture();
+        }
+        var configuration = consortiumConfiguration.get();
+        return settingService.getSettingByKey(SettingKey.CENTRAL_ORDERING_ENABLED, headers, getContext())
```
SerhiiNosko, line 99: we also need to use cache here, this setting never changes, so by using cache we can reduce lot of calls to DB and so these sql connections from pool can be used for other important work

src/test/java/org/folio/event/handler/InventoryCreateAsyncRecordHandlerTest.java
```diff
@@ -95,12 +112,51 @@ var record = createKafkaRecord(eventObject, DIKU_TENANT);
     });
   }
 
+  @Test
+  void positive_shouldSkipInventoryCreateEventIfCentralOrderingIsDisabled() {
+    var eventObject = createResourceEvent(DIKU_TENANT, CREATE);
+    var record = createKafkaRecord(eventObject, DIKU_TENANT);
+    doReturn(Future.succeededFuture(Response.ok(createSettingCollection()).build()))
```
azizbekxm, line 119: we can move createSettingCollection() to variable 
var emptySetting = new SettingCollection(); it is more easier to understand

azizbekxm (approved):

src/main/java/org/folio/services/setting/SettingService.java
Saba-Zedginidze-EPAM: Simplified a bit

src/test/java/org/folio/event/handler/InventoryCreateAsyncRecordHandlerTest.java
Saba-Zedginidze-EPAM: Replaced, thanks

src/main/java/org/folio/event/handler/InventoryCreateAsyncRecordHandler.java
Saba-Zedginidze-EPAM: Added cache here aswell

SerhiiNosko (approved):

PR #437, opened 2024-10-04
[MODORDERS-1192]. Add consumer to handle Holding Update event to update all other PO Line holdings to the same instance

BKadirkhodjaev:
## Purpose

## Approach

- Updates POL `searchLocationIds` with holding record `permanentLocationId` (does not remove old `searchLocationId`)
- Updates POL `instanceId` with holding record `instanceId`
- Updates each holdings record's `instanceId` in cascade, for every POL `holdingId` found in the `locations` array
  - Only if the `instanceId` was changed for at least one related holding record
  - Use batch endpoint `POST /holdings-storage/batch/synchronous?upsert=true`
- Add worker pool size configuration for each consumer verticle
- Test using unit tests


src/main/java/org/folio/services/lines/PoLinesService.java
```diff
@@ -221,6 +223,12 @@ public Future<List<PoLine>> getPoLinesByHoldingId(String holdingId, Conn conn) {
     return getPoLinesByField(criterion, conn);
   }
 
+  public Future<List<PoLine>> getPoLinesByCqlQuery(String query, Conn conn) throws FieldException {
```
BKadirkhodjaev, line 226: Added to as a potential replacement of `getPoLinesByHoldingId()`, SQL with `location.holdingId` does not return any rows

src/main/java/org/folio/verticles/AbstractConsumersVerticle.java
```diff
@@ -84,12 +94,18 @@ protected Future<KafkaConsumerWrapper<String, String>> createConsumer(T eventTyp
       .processRecordErrorHandler((t, r) -> log.error("Failed to process event: {}", r, t))
       .build();
 
-    log.info("createConsumer:: moduleId={}", MODULE_ID);
-    return consumerWrapper.start(handler, MODULE_ID)
+    String consumerName;
+    if (Objects.nonNull(methodName)) {
```
BKadirkhodjaev, line 98: Added to produce a unique consumer group to avoid work stealing on `inventory.holdings-record` by `CREATE` consumer, `mod-orders` has Data Import consumer which use this consumer name pattern

src/main/java/org/folio/event/util/KafkaEventUtil.java
```diff
@@ -27,4 +28,9 @@ public static String extractValueFromHeaders(List<KafkaHeader> headers, String k
       .orElse(null);
   }
 
+  public static Map<String, String> getHeaderMap(List<KafkaHeader> headers) {
```
BKadirkhodjaev, line 31: Changed to `TreeMap` to support case insensitive key, previous implementation return null of `X-Okapi-Tenant` header key (Kafka headers use lowercased header keys). For example: 

`{
	"folio.tenantId": "diku",
	"x-okapi-token": "eyJhbGciOiJIUzI1NiJ9.eyJzdWIiOiJVTkRFRklORURfVVNFUl9fMTcyLjE3LjAuMTo1NTE0NF9fMjAyNC0xMC0wN1QwNDo1NzoxMi41NzgrMDAwMCIsInR5cGUiOiJtb2R1bGUiLCJleHRyYV9wZXJtaXNzaW9ucyI6WyJTWVMjbW9kLWludmVudG9yeS1zdG9yYWdlLTI3LjIuMC1TTkFQU0hPVC4xMDMzIy9fL3RlbmFudCNbUE9TVF0iXSwidGVuYW50IjoiZGlrdSIsIm1vZHVsZSI6Im1vZC1pbnZlbnRvcnktc3RvcmFnZS0yNy4yLjAtU05BUFNIT1QuMTAzMyJ9.Bd7trhMwtdWN2CUnT9mBZ5p59EYOudlG1QPaYeE3e6k",
	"x-okapi-url": "http://host.docker.internal:9130",
	"x-okapi-tenant": "diku",
	"x-okapi-request-id": "936415/proxy;487539/tenant"`

src/main/java/org/folio/event/handler/HoldingUpdateAsyncRecordHandler.java
```diff
@@ -0,0 +1,193 @@
...
+    holdingIds.forEach(holdingId -> poLineFutures.add(updateAdjacentHolding(okapiUrl, requestHeaders, holdingId, newInstanceId)));
+    return GenericCompositeFuture.all(poLineFutures)
+      .onComplete(asyncResult -> log.info("updateAdjacentHoldingsWithNewInstanceId:: Updated adjacent holdings, size: {}", holdingIds.size()))
+      .mapEmpty();
+  }
+
+  private Future<Void> updateAdjacentHolding(String okapiUrl, MultiMap requestHeaders, String holdingId, String newInstanceId) {
+    log.info("updateAdjacentHolding:: Updating an adjacent holding, holdingId: {}, instanceId: {}", holdingId, newInstanceId);
+    var absoluteUri = String.format(STORAGE_HOLDING_URL, okapiUrl, holdingId);
+    return webClient.requestAbs(HttpMethod.GET, absoluteUri)
```
SerhiiNosko, line 139: Use RestClient for doing such calls to keep abstaction like Service -> RestClient -> WebClient

src/main/java/org/folio/event/handler/HoldingUpdateAsyncRecordHandler.java
```diff
@@ -0,0 +1,193 @@
...
+                                                    String tenantId, String holdingId, Conn conn) throws FieldException {
+    return poLinesService.getPoLinesByCqlQuery(String.format(PO_LINE_LOCATIONS_HOLDING_ID_CQL, holdingId), conn)
+      .compose(poLines -> updatePoLines(resourceEvent, tenantId, holdingId, poLines, conn))
+      .compose(poLines -> auditOutboxService.saveOrderLinesOutboxLogs(conn, poLines, OrderLineAuditEvent.Action.EDIT, headers).map(poLines));
+  }
+
+  private Future<List<PoLine>> updatePoLines(ResourceEvent resourceEvent, String tenantId,
+                                             String holdingId, List<PoLine> poLines, Conn conn) {
+    if (CollectionUtils.isEmpty(poLines)) {
+      log.warn("updatePoLines:: No POLs were found for holding to update, holdingId: {}", holdingId);
```
SerhiiNosko, line 91: it should be debug or info level, it is common situation

src/main/java/org/folio/event/handler/HoldingUpdateAsyncRecordHandler.java
```diff
@@ -0,0 +1,193 @@
...
+      .filter(Objects::nonNull)
+      .filter(holdingId -> !holdingId.equals(excludingHoldingId))
+      .distinct()
+      .toList();
+  }
+
+  private Future<Void> updateAdjacentHoldingsWithNewInstanceId(ResourceEvent resourceEvent, Map<String, String> headers,
+                                                               List<String> holdingIds) {
+    if (CollectionUtils.isEmpty(holdingIds)) {
+      log.warn("updateAdjacentHoldingsWithNewInstanceId:: No adjacent holdings were found to update, ignoring update");
```
SerhiiNosko, line 123: the same about log level

src/main/java/org/folio/event/handler/HoldingUpdateAsyncRecordHandler.java
```diff
@@ -0,0 +1,193 @@
...
+    return dbClient.getPgClient()
+      .withTrans(conn -> {
+        try {
+          return processPoLinesUpdate(resourceEvent, headers, tenantId, holdingId, conn)
+            .compose(poLines -> {
+              var adjacentHoldingIds = extractDistinctAdjacentHoldingsToUpdate(holdingId, poLines);
+              return updateAdjacentHoldingsWithNewInstanceId(resourceEvent, headers, adjacentHoldingIds);
+            });
+        } catch (FieldException e) {
+          throw new IllegalStateException("Produced and invalid CQL wrapper", e);
```
SerhiiNosko, line 73: do you think we need this catch? there is can be sitatuations when we could not control cql wrappers and want to logs exception about this with rethrowing?

BKadirkhodjaev: > do you think we need this catch? there is can be sitatuations when we could not control cql wrappers and want to logs exception about this with rethrowing?

Yea its not needed, I will make it a runtime exception so we don't need to handle or rethrow it

src/main/java/org/folio/event/handler/HoldingUpdateAsyncRecordHandler.java
```diff
@@ -0,0 +1,193 @@
...
+
+  private boolean updatePoLinesSearchLocationIds(ResourceEvent resourceEvent, List<PoLine> poLines) {
+    var pair = getIdPair(resourceEvent, PERMANENT_LOCATION_ID);
+    if (pair.getLeft().equals(pair.getRight())) {
+      log.warn("updatePoLinesSearchLocationIds:: No search location id was changed (ids are the same), ignoring update");
+      return false;
+    }
+    for (var poline : poLines) {
+      var searchLocationIds = poline.getSearchLocationIds();
+      if (searchLocationIds.remove(pair.getLeft())) {
```
SerhiiNosko, line 174: need to think can we remove searchLocationId, because there can be case when this searchLocationId is used also by other holding connected to the same po line

src/main/java/org/folio/event/handler/HoldingUpdateAsyncRecordHandler.java
SerhiiNosko: if it checked exception we also can consider lombock to do the same to simplify code

BKadirkhodjaev: > if it checked exception we also can consider lombock to do the same to simplify code

`@SneakyThrows`?

src/main/java/org/folio/event/handler/HoldingUpdateAsyncRecordHandler.java
```diff
@@ -0,0 +1,193 @@
...
+import static org.folio.event.InventoryEventType.INVENTORY_HOLDING_UPDATE;
+
+@Log4j2
+public class HoldingUpdateAsyncRecordHandler extends InventoryUpdateAsyncRecordHandler {
+
+  public static final String ID = "id";
+  public static final String INSTANCE_ID = "instanceId";
+  public static final String PERMANENT_LOCATION_ID = "permanentLocationId";
+  public static final String PO_LINE_LOCATIONS_HOLDING_ID_CQL = "locations==*%s*";
+  public static final String STORAGE_HOLDING_URL = "%s/holdings-storage/holdings/%s";
```
SerhiiNosko, line 43: and using bulk endpoint to update holdings was not possible?

BKadirkhodjaev: > and using bulk endpoint to update holdings was not possible?

I will try it again

src/main/java/org/folio/rest/impl/InitAPIs.java
```diff
@@ -78,11 +67,38 @@ public void init(Vertx vertx, Context context, Handler<AsyncResult<Boolean>> res
...
+  }
+
   private Future<?> deployKafkaConsumersVerticles(Vertx vertx) {
-    AbstractApplicationContext springContext = vertx.getOrCreateContext().get("springContext");
+    var springContext = (AbstractApplicationContext) vertx.getOrCreateContext().get(SPRING_CONTEXT_KEY);
     var consumers = List.of(
       deployVerticle("inventory-item-consumers", itemConsumersVerticleNumber, InventoryItemConsumersVerticle.class, vertx, springContext),
-      deployVerticle("inventory-holding-consumers", holdingConsumersVerticleNumber, InventoryHoldingConsumersVerticle.class, vertx, springContext),
+      deployVerticle("inventory-holding-create-consumers", holdingCreateConsumersVerticleNumber, InventoryHoldingCreateConsumersVerticle.class, vertx, springContext),
+      deployVerticle("inventory-holding-update-consumers", holdingUpdateConsumersVerticleNumber, InventoryHoldingUpdateConsumersVerticle.class, vertx, springContext),
```
SerhiiNosko, line 101: a bit worry about introducing new verticle. So we have worker verticle that will not be mixed with main event bus. But new worker thread has by default 20 threads in the pool as I remember(it is configurable). It can be so many for all kafka topics.
Initially thought that it would be good to have 1 verticle per item CRUD and 1 verticle per holding CRUD

BKadirkhodjaev: > a bit worry about introducing new verticle. So we have worker verticle that will not be mixed with main event bus. But new worker thread has by default 20 threads in the pool as I remember(it is configurable). It can be so many for all kafka topics. Initially thought that it would be good to have 1 verticle per item CRUD and 1 verticle per holding CRUD

How about reducing the number of worker threads?

src/main/java/org/folio/services/inventory/InventoryUpdateService.java
```diff
@@ -0,0 +1,65 @@
...
+    if (Objects.isNull(holdingIds) || CollectionUtils.isEmpty(holdingIds)) {
+      log.info("batchUpdateAdjacentHoldingsWithNewInstanceId:: No adjacent holdings were found to update, ignoring update");
+      return Future.succeededFuture();
+    }
+    return batchUpdateAdjacentHoldings(holdingIds, holder.getInstanceId(), requestContext)
+      .onComplete(asyncResult -> log.info("batchUpdateAdjacentHoldingsWithNewInstanceId:: Updated adjacent holdings, size: {}", holdingIds.size()))
+      .mapEmpty();
+  }
+
+  private Future<Void> batchUpdateAdjacentHoldings(List<String> holdingIds, String newInstanceId, RequestContext requestContext) {
```
BKadirkhodjaev, line 48: Uses the new batch endpoint

src/main/java/org/folio/event/handler/HoldingUpdateAsyncRecordHandler.java
SerhiiNosko: looks good

src/main/java/org/folio/rest/impl/InitAPIs.java
SerhiiNosko: solved by adding configurable worker pool size

src/main/java/org/folio/event/util/KafkaEventUtil.java
```diff
@@ -27,4 +28,9 @@ public static String extractValueFromHeaders(List<KafkaHeader> headers, String k
...
+    var caseInsensitiveMap = new TreeMap<String, String>(String.CASE_INSENSITIVE_ORDER);
+    headers.forEach(header -> caseInsensitiveMap.put(header.key(), header.value().toString()));
```
azizbekxm, line 33: 1. in case of header.value() is null 
headers.forEach(header -> {
        var key = header.key();
        var value = header.value();
        caseInsensitiveMap.put(key, value != null ? value.toString() : "");
    });
    2. We can also use this option
     var headers = new CaseInsensitiveMap<>(kafkaHeadersToMap(kafkaConsumerRecord.headers())); example:https://github.com/folio-org/mod-orders-storage/blob/d14a5447ca8e1f38d24c3cd4df9b20b2231dc13e/src/main/java/org/folio/event/handler/InventoryCreateAsyncRecordHandler.java#L70
     
     kafkaHeadersToMap(existing method in dependency)
     because TreeMap have to main order and access to field is o(logn) while CaseInsentivemap is litte more performance for get operation.

src/main/java/org/folio/rest/impl/InitAPIs.java
```diff
@@ -78,21 +75,47 @@ public void init(Vertx vertx, Context context, Handler<AsyncResult<Boolean>> res
...
     return GenericCompositeFuture.all(consumers)
       .onSuccess(future -> log.info("deployKafkaConsumersVerticles:: All {} consumer(s) was successfully started", consumers.size()))
       .onFailure(exception -> log.error("Failed to start consumers", exception));
   }
 
-  private <V extends Verticle> Future<String> deployVerticle(String poolName, int instancesNumber, Class<V> consumerClass, Vertx vertx, AbstractApplicationContext springContext) {
-    return vertx.deployVerticle(() -> springContext.getBean(consumerClass),
-      new DeploymentOptions().setThreadingModel(ThreadingModel.WORKER).setWorkerPoolName(poolName).setInstances(instancesNumber));
+  private <V extends Verticle> Future<String> deployVerticle(String poolName, int instancesNumber, int poolSize, Class<V> consumerClass, Vertx vertx, AbstractApplicationContext springContext) {
+    var deploymentOptions = new DeploymentOptions().setThreadingModel(ThreadingModel.WORKER).setWorkerPoolName(poolName).setInstances(instancesNumber).setWorkerPoolSize(poolSize);
```
azizbekxm, line 118: seems some lines is too long can be moved to next line. please look at other places as well. long lines is hard to read in laptop screen or dividied screen in monitor

src/main/java/org/folio/event/handler/HoldingUpdateAsyncRecordHandler.java
BKadirkhodjaev: Removed id deletion as planned

azizbekxm (approved):

PR #441, opened 2024-10-11
[MODORDERS-1192-3]. Add ECS support to holdings-record update consumer

BKadirkhodjaev:
## Purpose

## Approach

- Add ECS support to `holdings-record` update consumer

## Related issues

- <https://github.com/folio-org/mod-orders-storage/pull/437>
- <https://github.com/folio-org/mod-orders-storage/pull/440>


src/main/java/org/folio/event/handler/HoldingUpdateAsyncRecordHandler.java
```diff
@@ -84,7 +84,10 @@ private Future<List<PoLine>> updatePoLines(InventoryUpdateHolder holder, List<Po
       log.info("updatePoLines:: No POLs were updated for holding, holdingId: {}, POLs retrieved: {}", holder.getHoldingId(), poLines.size());
       return Future.succeededFuture(List.of());
     }
-    return poLinesService.updatePoLines(poLines, conn, holder.getTenantId())
+    // Must pass the correct active tenantId to resolve which schema name is to be used in the poLine update
+    // in a non-ecs environment the active tenantId is the same as the event tenantId
+    // in an ecs environment the active tenantId is the central tenantId
```
SerhiiNosko, line 89: minor: can be added that only with central ordering feature enabled. Because we can have libraries that use ECS and have orders in each separate tenant

src/main/java/org/folio/services/consortium/ConsortiumConfigurationService.java
```diff
@@ -3,43 +3,76 @@
...
       .executor(task -> Vertx.currentContext().runOnContext(v -> task.run()))
       .buildAsync();
   }
 
+  public Future<String> getCentralTenantId(Context context, Map<String, String> headers) {
+    var requestContext = new RequestContext(context, headers);
+    return getConsortiumConfiguration(requestContext)
+      .compose(consortiumConfiguration -> {
+        if (consortiumConfiguration.isEmpty()) {
+          log.info("getCentralTenantId:: No settings were found");
```
SerhiiNosko, line 56: we will have this log for each non ecs event processing, can be debug or removed

SerhiiNosko (approved):

src/main/java/org/folio/event/handler/InventoryUpdateAsyncRecordHandler.java
```diff
@@ -74,6 +80,12 @@ private static InventoryUpdateHolder createInventoryUpdateHolder(ResourceEvent r
       .build();
   }
 
+  // Prepare a DB client for particular tenant
+  private Future<Void> prepareProcessInventoryUpdateEvent(InventoryUpdateHolder holder, String centralTenantId) {
+    holder.setCentralTenantId(centralTenantId);
+    return processInventoryUpdateEvent(holder, createDBClient(holder.getActiveTenantId()));
+  }
```
azizbekxm, line 87: interesting case, what if we have ecs, but consortium setting disabled. if kafka send member tenant, and based on current approach it will try to search member tenant because central tenant is null. In the end dbClient can fint from member because it is localted in central tenants. For Create Consumer we didn't proceess event

BKadirkhodjaev: > interesting case, what if we have ecs, but consortium setting disabled. if kafka send member tenant, and based on current approach it will try to search member tenant because central tenant is null. In the end dbClient can fint from member because it is localted in central tenants. For Create Consumer we didn't proceess event

Yeah I see, the alternative is to throw an exception or stop processing this event by silently returning

SerhiiNosko: Disabling this setting is not possible, at least for now, disabling is also not allowed from UI

azizbekxm (approved):

PR #449, opened 2024-10-22
[MODORDSTOR-416-5]. Add additional consumer improvements

BKadirkhodjaev:
## Purpose

- <https://folio-org.atlassian.net/browse/MODORDSTOR-416>
- <https://folio-org.atlassian.net/browse/MODORDSTOR-419>
- Add changes improving item and holding consumers
- Fix failing outbox logs which trigger a transaction rollback

## Approach

- Add additional NPE protection to POL update
- Add additional NPE protection to piece update
- Add another transaction for outbox logging using the event tenant id
- Update JsonObject deserialization for both POL and piece batch query methods
- Move event verification before fetch central tenant

### DIFF
--- a/src/main/java/org/folio/event/handler/HoldingCreateAsyncRecordHandler.java
+++ b/src/main/java/org/folio/event/handler/HoldingCreateAsyncRecordHandler.java
@@ -14,9 +14,7 @@ import org.folio.event.dto.ResourceEvent;
 import org.folio.event.service.AuditOutboxService;
 import org.folio.okapi.common.GenericCompositeFuture;
 import org.folio.rest.jaxrs.model.Location;
-import org.folio.rest.jaxrs.model.OrderLineAuditEvent;
 import org.folio.rest.jaxrs.model.Piece;
-import org.folio.rest.jaxrs.model.PieceAuditEvent;
 import org.folio.rest.jaxrs.model.PoLine;
 import org.folio.rest.persist.Conn;
 import org.folio.rest.persist.DBClient;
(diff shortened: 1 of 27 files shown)


azizbekxm:
image

Saba-Zedginidze-EPAM (approved):

src/main/java/org/folio/event/handler/HoldingCreateAsyncRecordHandler.java
```diff
@@ -56,32 +56,32 @@ protected Future<Void> processInventoryCreationEvent(ResourceEvent resourceEvent
     var permanentLocationId = holdingObject.getString(PERMANENT_LOCATION_ID.getValue());
     var tenantIdFromEvent = resourceEvent.getTenant();
     return dbClient.getPgClient()
-      .withTrans(conn -> {
+      .withTrans(conn -> createDBClient(tenantIdFromEvent).getPgClient().withTrans(connAudit -> {
```
SerhiiNosko, line 59: if it 2 different sql connection - the same transaction is not possible that can use 2 connections.

SerhiiNosko: it seams need some refactoring to save audit logs separately in scope of another connection without transaction.
If it is complex lets merge all other optimization at least to pass karate tests, and refactoring in scope of another PR

azizbekxm:
image

src/main/java/org/folio/event/handler/HoldingCreateAsyncRecordHandler.java
BKadirkhodjaev: > it seams need some refactoring to save audit logs separately in scope of another connection without transaction. If it is complex lets merge all other optimization at least to pass karate tests, and refactoring in scope of another PR

Will try refactoring now

src/main/java/org/folio/event/handler/HoldingUpdateAsyncRecordHandler.java
```diff
@@ -76,7 +81,7 @@ private Future<List<PoLine>> processPoLinesUpdate(HoldingEventHolder holder, Con
       })
       .compose(poLines -> updatePoLines(holder, poLines, conn))
       .compose(poLines -> updateTitles(holder, poLines, conn).map(poLines))
-      .compose(poLines -> auditOutboxService.saveOrderLinesOutboxLogs(conn, poLines, OrderLineAuditEvent.Action.EDIT, holder.getHeaders()).map(poLines));
+      .onSuccess(poLines -> auditOutboxService.saveOrderLinesOutboxLogs(conn, poLines, OrderLineAuditEvent.Action.EDIT, holder.getHeaders()).map(poLines));
```
SerhiiNosko, line 84: can we remain compose here?

src/main/java/org/folio/util/DbUtils.java
```diff
@@ -65,9 +67,10 @@ public static <T> List<T> getRowSetAsList(RowSet<Row> rowSet, Class<T> entityCla
   }
 
   public static <T> T convertResponseToEntity(Response response, Class<T> entityClass) {
-    return JsonObject.mapFrom(response.getEntity()).mapTo(entityClass);
+    try {
+      return JsonObject.mapFrom(response.getEntity()).mapTo(entityClass);
+    } catch (RuntimeException e) {
+      throw new IllegalStateException(String.format("Cannot convert response '%s' to entity '%s'", response.getEntity(), entityClass.getName()));
```
SerhiiNosko, line 73: better to add e.getMessage()

src/main/java/org/folio/event/handler/HoldingUpdateAsyncRecordHandler.java
BKadirkhodjaev: > can we remain compose here?

You mean to have 2 compose in a TX for PoLine and Piece, then followed by that 1 for PoLine and 1 for Piece outbox, 4 compose commands chained?

BKadirkhodjaev: > can we remain compose here?

I see what you mean, change onSuccess to compose, yeah it can be done

SerhiiNosko: yes, 4 compose like this
image

SerhiiNosko: because if we remain this logic in onSuccess we will have success future as result so will not know about failure if could not write to audit log

SerhiiNosko (approved):

PR #457, opened 2024-11-11
[MODORDSTOR-431]. ECS | Update search locations for POL created in Central tenant after updating the permanent location of the holding from Member tenant related to the POL

BKadirkhodjaev:
## Purpose

- <https://folio-org.atlassian.net/browse/MODORDSTOR-431>

## Approach

- Add an alternative mode to update POL search location ids using central tenant id
  - Triggers on no update of POLs in using a member tenant id
- Tested with unit tests and in 3 different modes manually: Single-tenant, ECS Central tenant and ECS member tenant

### DIFF
--- a/src/main/java/org/folio/event/dto/HoldingUpdate.java
+++ b/src/main/java/org/folio/event/dto/HoldingUpdate.java
@@ -1,0 +1,23 @@
+package org.folio.event.dto;
+
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+import org.folio.rest.jaxrs.model.PoLine;
+
+import java.util.List;
+
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class HoldingUpdate {
+
+  private int affectedRows;
+  private boolean isInstanceIdUpdated;
+  private boolean isSearchLocationIdsUpdated;
+  private List<PoLine> poLineWithUpdatedInstanceId;
+  private List<PoLine> poLineWithUpdatedSearchLocationIds;
+  private List<String> adjacentHoldingIds;
+}
--- a/src/main/java/org/folio/event/dto/HoldingUpdate.java
+++ b/src/main/java/org/folio/event/dto/HoldingUpdate.java
@@ -17,7 +17,7 @@ public class HoldingUpdate {
   private int affectedRows;
   private boolean isInstanceIdUpdated;
   private boolean isSearchLocationIdsUpdated;
-  private List<PoLine> poLineWithUpdatedInstanceId;
-  private List<PoLine> poLineWithUpdatedSearchLocationIds;
+  private List<PoLine> poLinesWithUpdatedInstanceId;
+  private List<PoLine> poLinesWithUpdatedSearchLocationIds;
   private List<String> adjacentHoldingIds;
 }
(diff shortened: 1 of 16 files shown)


src/main/java/org/folio/event/handler/HoldingUpdateAsyncRecordHandler.java
```diff
@@ -57,7 +58,18 @@ protected Future<Void> processInventoryUpdateEvent(ResourceEvent resourceEvent,
       log.info("processInventoryUpdateEvent:: No instance id or search location ids to update in holding '{}', ignoring update", holder.getHoldingId());
       return Future.succeededFuture();
     }
-    return processHoldingUpdateEvent(holder);
+    return consortiumConfigurationService.getCentralTenantId(getContext(), headers)
+      .compose(centralTenantId -> {
+        holder.setCentralTenantId(centralTenantId);
+        return processHoldingUpdateEvent(holder)
+          .compose(poLines -> {
+            if (Boolean.FALSE.equals(holder.isPoLinesUpdated()) && Objects.nonNull(centralTenantId)) {
```
SerhiiNosko, line 66: if possible to return future with booelan value for processHoldingUpdateEvent() it would be easier and eliminate poLinesUpdated property in holder

SerhiiNosko: okay, if we could not rely on return value, lets remain as it

BKadirkhodjaev: > okay, if we could not rely on return value, lets remain as it (its like contracts for updating rows in DB, if some rows updated Spring Data returns true or false otherwsie)

I will check the affected rows after POL update tomorrow

azizbekxm (approved):

src/main/java/org/folio/event/dto/HoldingUpdate.java
```diff
@@ -0,0 +1,23 @@
...
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class HoldingUpdate {
+
+  private int affectedRows;
+  private boolean isInstanceIdUpdated;
+  private boolean isSearchLocationIdsUpdated;
+  private List<PoLine> poLineWithUpdatedInstanceId;
```
SerhiiNosko, line 20: poLinesWithUpdatedInstanceId - minor renaming

SerhiiNosko (approved):

src/main/java/org/folio/event/handler/HoldingCreateAsyncRecordHandler.java
```diff
@@ -58,7 +58,7 @@ protected Future<Void> processInventoryCreationEvent(ResourceEvent resourceEvent
     var tenantIdFromEvent = resourceEvent.getTenant();
     return dbClient.getPgClient()
       .withTrans(conn -> {
-        var updatedHeaders = HeaderUtils.copyHeadersAndUpdatedTenant(centralTenantId, headers);
+        var updatedHeaders = HeaderUtils.prepareHeaderForTenant(centralTenantId, headers);
```
BKadirkhodjaev, line 61: Note to future me:

- Kafka headers are not case-insensitive: 
image

Thus any implementation or manipulation of Kafka headers must be performed using a case-insensitive Map implementation, such as `org.apache.commons.collections4.map.CaseInsensitiveMap<K, V>`

PR #489, opened 2025-05-16
[MODORDSTOR-458]. Invalid reference is displayed in PO line after changing instance connection for the P/E mix order

BKadirkhodjaev:
## Purpose

- <https://folio-org.atlassian.net/browse/MODORDSTOR-458>

## Approach

- TODO

### DIFF
--- a/src/main/java/org/folio/orders/lines/update/instance/WithHoldingOrderLineUpdateInstanceStrategy.java
+++ b/src/main/java/org/folio/orders/lines/update/instance/WithHoldingOrderLineUpdateInstanceStrategy.java
@@ -3,6 +3,7 @@ package org.folio.orders.lines.update.instance;
 import io.vertx.core.Promise;
 import io.vertx.ext.web.handler.HttpException;
 import lombok.RequiredArgsConstructor;
+import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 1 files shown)


src/main/java/org/folio/orders/lines/update/instance/WithHoldingOrderLineUpdateInstanceStrategy.java
```diff
@@ -83,11 +81,25 @@ private Future<Tx<PoLine>> updateHoldings(Tx<PoLine> poLineTx, ReplaceInstanceRe
...
           }
         });
+    } else {
+      log.info("Holding does not require an update");
+      return Future.succeededFuture(poLineTx);
     }
   }
 
   private boolean isUpdatedHolding(List<Holding> holdings) {
-    return holdings.stream().noneMatch(holding -> holding.getFromHoldingId().equals(holding.getToHoldingId()));
```
BKadirkhodjaev, line 90: This noneMatch was short-circuiting if holdingId was the same, i.e. if we have fromHoldingId == toHoldingId,  isUpdatedHolding would return early with false

SerhiiNosko (approved):

PR #492, opened 2025-06-02
Add categoryIds to the reference data

BKadirkhodjaev:
## Purpose

- Add categoryIds to the reference data

## Approach

- Add a new categoryIds field to the reference data so that Order Templates can be created on Tenant Entitlement

### DIFF
--- a/src/main/resources/data/order-templates/amazon_book_orders.json
+++ b/src/main/resources/data/order-templates/amazon_book_orders.json
@@ -33,5 +33,6 @@
       "amazon"
     ]
   },
-  "vendor": "e0fb5df2-cdf1-11e8-a8d5-f2801f1b9fd1"
+  "vendor": "e0fb5df2-cdf1-11e8-a8d5-f2801f1b9fd1",
+  "categoryIds": []
 }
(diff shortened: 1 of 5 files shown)


src/main/resources/data/order-templates/amazon_book_orders.json
```diff
@@ -33,5 +33,6 @@
       "amazon"
     ]
   },
-  "vendor": "e0fb5df2-cdf1-11e8-a8d5-f2801f1b9fd1"
+  "vendor": "e0fb5df2-cdf1-11e8-a8d5-f2801f1b9fd1",
+  "categoryIds": []
```
SerhiiNosko, line 37: does it means that in scope of Trillium we need to create new migration script to update all existing order templates, because this functionality also was introduced in Trillium?

BKadirkhodjaev: > does it means that in scope of Trillium we need to create new migration script to update all existing order templates, because this functionality also was introduced in Trillium?

I think this may be needed, although I couldn't catch the exact error message on reference data loading on tenant entitlement

PR #493, opened 2025-06-03
[MODORDERS-1278]. Display the fiscal year in which the Order was opened based on "Date opened"

BKadirkhodjaev:
## Purpose

- <https://folio-org.atlassian.net/browse/MODORDERS-1278>

## Approach

- TODO


SerhiiNosko:
I remember I asked about migration script on refinement and answer was that it is needed. In any case we can merge this PR as it and for migration script create another PR, I am not sure that we can populate 'openedById' but also need to check can we populate 'fiscalYearId'

SerhiiNosko (approved):

PR #497, opened 2025-06-13
[MODORDSTOR-433] Replace mod-configuration to retrieve current time zone in Claiming job

Saba-Zedginidze-EPAM:
## Purpose
[[MODORDSTOR-433] Replace mod-configuration to retrieve current time zone in Claiming job](https://folio-org.atlassian.net/browse/MODORDSTOR-433)

## Approach
- Add a new CommonSettingsService with javadocs to differentiate it form SettingsService
- Replace existing TenantLocaleSettingsService with new one
- Update and add new tests


SerhiiNosko (approved):

src/test/java/org/folio/services/setting/CommonSettingsServiceTest.java
```diff
@@ -0,0 +1,95 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+import io.vertx.core.Future;
+import io.vertx.core.json.JsonObject;
+
+@CopilotGenerated(partiallyGenerated = true)
```
BKadirkhodjaev, line 29: Please also add why model you were using for this tests, I think we already have a new `model` field on this annotation

BKadirkhodjaev (approved):

PR #499, opened 2025-06-30
[MODORDSTOR-469] Update `folio-custom-fields` dependency

alb3rtino:
## Purpose
https://folio-org.atlassian.net/browse/MODORDSTOR-469

## Approach
Update `folio-custom-fields` to latest version. Fixes https://folio-org.atlassian.net/browse/FCFIELDS-68.


SerhiiNosko:
Hi @alb3rtino , is this reproducible also on Sunflower Bugfest? If yes then we can request approvals for CSP for Sunflower and you can release it

SerhiiNosko:
okay, it is nice that no hot fixes are required here

PR #500, opened 2025-07-01
[MODORDERS-1321]. Add 'Suppress from discovery' toggle for instance interaction

BKadirkhodjaev:
## Purpose

- <https://folio-org.atlassian.net/browse/MODORDERS-1321>

## Approach

- Update acq-models


SerhiiNosko:
about migration script to make default value 'false' for existing records I think we can skip it. So for processing this flag we should consider that both null and false values are possible.

SerhiiNosko (approved):

PR #648, opened 2026-07-17
[MODORDSTOR-524] Title name is not updated after changing instance connection 

ykashtalian:
### **Purpose**
[MODORDSTOR-524](https://folio-org.atlassian.net/browse/MODORDSTOR-524) - Title name is not updated after changing instance connection.

This PR fixes the deadlocks caused by the same two concurrent transactions as the "Title name is not updated after changing instance connection" issue. The deadlocks were found while trying to reproduce that issue.

### **Approach**
ItemUpdateAsyncRecordHandler now locks po_line first, like the PATCH does.

### **Pre-Review Checklist**

- [ ] **Self-reviewed Code** — Reviewed code for issues, unnecessary parts, and overall quality.
- [ ] **Change Notes** — NEWS.md updated with clear description and issue key.
- [ ] **Testing** — Confirmed changes were tested locally or on dev environment.
- [ ] **Logging**: Confirmed that logging is appropriately handled.
- [ ] **Breaking Changes** — Handled all required actions if changes affect API, DB, or interface versions.
  - [ ] API/schema changes
  - [ ] Interface version updates
  - [ ] DB schema changes / migration scripts
- [ ] **New Properties / Environment Variables** — Updated README.md if new configs were added.

### DIFF
--- a/src/main/java/org/folio/event/handler/ItemUpdateAsyncRecordHandler.java
+++ b/src/main/java/org/folio/event/handler/ItemUpdateAsyncRecordHandler.java
@@ -13,6 +13,7 @@ import io.vertx.core.Context;
 import io.vertx.core.Future;
 import io.vertx.core.Vertx;
 import lombok.extern.log4j.Log4j2;
+import one.util.streamex.StreamEx;
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.lang.BooleanUtils;
 import org.apache.commons.lang.ObjectUtils;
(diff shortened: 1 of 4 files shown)


src/main/java/org/folio/event/handler/ItemUpdateAsyncRecordHandler.java
```diff
@@ -66,12 +69,29 @@ private Future<Void> processItemUpdateEvent(ItemEventHolder holder) {
...
+   * order would cause a deadlock.
+   */
+  private Future<Void> lockPoLinesBeforePieces(ItemEventHolder holder, Conn conn) {
+    return pieceService.getPiecesByItemId(holder.getItemId(), conn)
+      .map(pieces -> filterPiecesToUpdate(holder, pieces).stream().map(Piece::getPoLineId).distinct().toList())
+      .compose(poLineIds -> {
+        if (CollectionUtils.isEmpty(poLineIds)) {
+          return Future.succeededFuture();
+        }
+        return poLinesService.getPoLinesByIdsForUpdate(poLineIds, holder.getOrderTenantId(), conn).mapEmpty();
```
SerhiiNosko: can we use these po lines in futher calls instead of returning null? In this case later we will not need to make an additional call

ykashtalian: You mean reuse it inside the ` processPoLinesUpdate` call?

ykashtalian: So, do you think that I should also add the lock on po lines inside `ItemCreateAsyncRecordHandler`?
Because in case we'll change the signature of `updatePoLineLocationData(List<String> poLineIds, ...` to receive PoLines instead of ids, we will need to fetch them in  `ItemCreateAsyncRecordHandler`.

Saba-Zedginidze-EPAM (approved):

SerhiiNosko (approved):

